PR #68, opened 2026-04-04
feat(dataapp-developer): v3.6.0 — deployment & design system fixes

ottomansky:
## Summary

- **Python version pinning**: `requires-python = ">=3.11,<3.13"` — prevents `uv` from downloading Python 3.14 which has no pre-built pandas/numpy wheels (builds from source = slow/fragile)
- **`.gitignore` fixes**: Allow standalone `node_modules/`, scope `dist/` and `build/` rules to `backend/` — fixes `Cannot find module 'next'` MODULE_NOT_FOUND at runtime
- **DataStatusBadge table links**: Correct Keboola Storage UI URL format (`{bucket}/overview/table/{name}/overview` instead of `/tables/{id}`)
- **ARIA accessibility**: Add `aria-pressed` and `aria-label` to FilterBar pill buttons and segmented controls
- **KAI_TOKEN**: Added to deployment secrets alongside KBC_TOKEN/KBC_URL
- **deployment.md rewrite**: Full alignment with canonical `dataapp-deployment` skill, documents all gotchas found during real deployment
- **Marketplace version**: bumped to 1.10.0, plugin to 3.6.0

## Test plan

- [ ] Deploy a new data app using the skill — verify `uv sync` uses Python 3.11 (no 3.14 download)
- [ ] Verify `.next/standalone/node_modules/next/dist/` is committed and app starts without MODULE_NOT_FOUND
- [ ] Verify DataStatusBadge table links open the correct Keboola Storage UI page
- [ ] Verify KAI chat works with KAI_TOKEN secret set


jordanrburger (commented):
## Deep Review — `feat(dataapp-developer): v3.6.0`

Ambitious and comprehensive rewrite. The 6-phase workflow, design system, and deployment docs are well-thought-out. However, there are several issues that would prevent successful end-to-end use. Organized by severity below.

---

## Critical (will break the skill or deployed apps)

### 1. `design-tokens.md` does not exist

Every reference file cross-links to it (`design-components.md:2`, `design-charts.md:2`, `design-advanced.md:2`). SKILL.md lists it as a required frontend agent read. `agent-prompts.md` makes it the **first file** the frontend agent reads. The file simply doesn't exist in the repo — the frontend agent will error or silently skip it, losing all color token definitions.

**Fix:** Create `references/design-tokens.md` with the color tokens, typography, number formatting, and z-layer definitions currently scattered across `globals.css` and `lib/constants.ts`.

### 2. Next.js standalone static files are never copied

`setup.sh` only runs `uv sync`. It does **not** copy `.next/static/` or `public/` alongside the standalone `server.js`. Without this, the deployed app loads HTML but all CSS, JS bundles, and images return 404. This is the #1 deployment gotcha with Next.js standalone output.

**Fix:** Add to `setup.sh`:
```bash
cp -r /app/frontend/.next/static /app/frontend/.next/standalone/.next/static
cp -r /app/frontend/public /app/frontend/.next/standalone/public
```

Or add nginx `location` blocks to serve static files directly.

### 3. Two incompatible `DataStatusBadge` implementations

- `design-components.md:175` — uses `useHealthCheck` + `/api/health` endpoint
- `design-advanced.md:403` — uses `useDataStatus` + `/api/data-status` endpoint

Different hooks, different backend contracts, different UI. The starter template's `HealthResponse` type (`lib/types.ts:8-10`) satisfies neither. An agent reading both files will produce broken code.

**Fix:** Pick one implementation. Mark `design-advanced.md`'s version as "alternative/enhanced" with an explicit note, or remove it.

### 4. `lucide-react@^0.500.0` does not exist

`frontend/package.json:21` — this version hasn't been released. `npm install` will fail.

**Fix:** Pin to an actual release (e.g., `^0.460.0` or whatever is current).

### 5. `globals.css` font tokens are circular self-references

Lines 25-26: `--font-sans: var(--font-sans)` resolves to `initial` (empty) in browsers. Tailwind v4 needs `@theme inline` for pass-through references.

**Fix:**
```css
@theme inline {
  --font-sans: var(--font-sans);
  --font-mono: var(--font-mono);
}
```

### 6. `echarts-for-react@^3` is incompatible with React 19

Peer dependency on React 16/17/18. Will produce install warnings or runtime errors.

**Fix:** Either remove `echarts-for-react` and use `echarts` directly, or find/pin a React 19-compatible version.

### 7. `deployment.md` pyproject.toml example is missing `[tool.setuptools.packages.find]`

Without `include = ["services*", "routers*"]`, `uv sync` won't locate the backend packages. App fails at import time.

### 8. `ai-assistant.md:156` — variable name bug

`buffer["events"]` should be `buf["events"]` — the variable is named `buf` at line 144. `NameError` at runtime in a "copy exactly" code snippet.

---

## High (confusing agent behavior or incorrect output)

### 9. `SKILL_REFS_PATH` timing is wrong

SKILL.md says "resolve before Phase 3" but uses it in Phase 0 (line 127) and Prerequisites (line 60). Should be "resolve at skill startup."

### 10. Phase 2 scaffold — template path is wrong

`SKILL_REFS_PATH` resolves to `references/` (anchored by `app-patterns.md`), but the template is at `templates/`. Agents need `$(dirname ${SKILL_REFS_PATH})/templates/...`.

### 11. Integration Check has wrong path

Phase 3 Integration Check item 6 greps `components/layout/NavTabs.tsx` — missing `frontend/` prefix. All other paths in that section include it.

### 12. No Stage B (post-deployment) verification agent

SKILL.md describes Stage A and Stage B verification, but `agent-prompts.md` only has a Phase 5 template for Stage A. Post-deployment verification has no agent prompt.

### 13. `mcp-setup.md` Steps 3 and 4 are inverted

Step 3 says "Once MCP is connected, run `get_project_info`" — but MCP isn't connected until Step 4 writes `.mcp.json` and the user restarts.

### 14. `app-patterns.md` CORS uses `allow_origins=["*"]`

Template correctly restricts to localhost. The doc's wildcard is a security regression if copied.

### 15. `.skeleton-shimmer` vs `.skeleton` class name mismatch

`design-advanced.md` uses `.skeleton-shimmer`. `globals.css` defines `.skeleton`. Different name, different gradients.

### 16. `formatCompact` is currency-only but used as general Y-axis formatter

Always prepends `$`. Will render `$1.2M` on user-count or percentage Y-axes.

### 17. `my-dashboards.md` — `C` variable used without import

`KaiTableChart.tsx` references `C` (defined in `chart-utils.ts`) but never imports it. `ReferenceError` at runtime.

### 18. `uv.lock` is missing from the template

`setup.sh` runs `uv sync` which needs a lockfile for reproducible installs.

### 19. Missing `KBC_TOKEN` fallback for Stage B when Stage A is skipped

If user picks "Deploy to Keboola" directly, `backend/.env` never gets written, but Stage B reads `KBC_TOKEN` from it.

---

## Medium (inconsistencies, misleading docs)

- `mcp-setup.md:155` lists forbidden `mcp__claude_ai_Keboola_GCP_EU__` as an example — contradicts its own rule
- `deployment.md` nginx `location = /` is missing WebSocket upgrade headers present in the template
- README dead link: references deleted `dataapp-deployment` skill path
- `deployment.md` pyproject.toml example missing `numpy>=2.0.0` pin
- `.orphaned_at` file committed — purpose unclear, possibly unintentional
- Recharts "setup pattern" code block violates the rules section below it (no `domain`, default tooltip)
- `validation.md` section numbering conflict: "3" (Recommended) vs "3b" (Required)
- `my-dashboards.md` hardcodes `#097cf7` in ~15 places instead of using `COLORS.brandPrimary`
- `providers.tsx` — `getQueryClient()` called at render time vs TanStack's recommended `useState` pattern
- `/api/chat` nginx block is dead config (no chat endpoint in template)
- `httpx` is a dependency but never imported
- `InfoPopover` in `design-components.md` has malformed JSX (mismatched closing tags)
- `EChartsInstance` is not an exported type from `echarts-for-react` v3

---

Overall this is solid work with a clear vision. Fixing the critical items (especially `design-tokens.md`, standalone static files, and the `DataStatusBadge` conflict) would make this shippable. Happy to re-review after fixes.

🤖 *Review generated with [Claude Code](https://claude.com/claude-code)*

jordanrburger:
@ottomansky - I'm closing this out. We need a more focused, manageable skill without all the extra design pieces.

PR #70, opened 2026-04-27
feat(sl-builder): add semantic-layer builder plugin

ottomansky:
## Summary

Adds the **sl-builder** plugin: an end-to-end skill that generates and pushes a semantic layer model to any Keboola project's metastore. Pipeline: schema discovery → SQL context → business questions → generate → validate (auto-fix + errors) → confirm & push.

Works with or without the Semantic Layer data app — talks to the metastore API directly with the project's storage token.

## What it does

- Lists projects from `~/Library/Application Support/keboola-agent-cli/config.json` and asks the user to pick one (never auto-selects)
- Detects existing models and offers partial update vs. new
- Filters tables down to fact/dim/L2 candidates, runs schema + SQL-transformation fetch in parallel via `kbagent`
- Generates `semantic-dataset`/`-metric`/`-relationship`/`-glossary`/`-constraint` payloads conforming to the metastore API
- Programmatic validation suite: auto-fixes PK/FK role, type mismatches vs Snowflake `column_details`, missing time-dimension joins; errors for phantom fields, dangling relationships, AGG-on-string, SUM-on-pct, name duplicates
- Pushes in parallel via `ThreadPoolExecutor`, with retry-on-transient-error
- Cleans up orphan items (filtered by `modelUUID`, since `?modelId` query param is unreliable)

## Operational guidance bundled in the skill

Beyond the build pipeline, the skill documents recurring traps that downstream pipeline consumers (Python apps materializing `DIM_METRIC`/`DIM_METRIC_THRESHOLD`, transformations joining on `CODE_METRIC`) hit in production:

- **Constraint severity encoding**: API only accepts `error`/`warning`/`info`, but health-band UIs typically need 4 levels — encode `_critical`/`_warning`/`_healthy`/`_review` in constraint name suffix
- **Constraint metric validation**: any name in `metrics[]` array that doesn't match an existing `semantic-metric` produces orphan FKs in downstream DIM tables
- **Collision-safe abbreviation generation**: naïve first-letters-of-words produces duplicates; provided generator uses full `seen` set instead of per-base counter to avoid `KDVPA`/`KDVPB` both renaming to `KDVP1`
- **Snowflake CTAS column-type cache trap**: once a destination column gets typed (e.g., `NUMBER(38,0)` from `COALESCE(..., 0)` with bare INT literal), the schema persists at the bucket level even after `delete-table`; reset with `kbagent storage delete-table --force` + `kbagent storage create-table --column "X:NUMBER(38,4)"`
- **Aliases block deletion**: `--force` cascade-deletes aliased tables in linked buckets
- **Cross-project source column drift**: same logical column may be `STRING` in one project, `NUMERIC` in another — `NULLIF(col,'')::NUMBER` works only for STRING; verify with `kbagent storage table-detail` per project
- **Promotion between projects**: `modelUUID` differs per project; export/replace UUID/import pattern provided

## Test plan

- [ ] Install plugin locally (clone fork or merge branch)
- [ ] Run `/sl-builder:semantic-layer` against a Keboola project with at least 2 fact tables
- [ ] Verify model is generated with non-empty datasets/metrics/glossary
- [ ] Confirm validation auto-fixes obvious type mismatches and rejects phantom fields
- [ ] Push to a test project, then re-run with partial update mode
- [ ] Verify constraints with name-suffix severity encoding round-trip correctly via metastore API

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://anthropic.com/claude-code/marketplace.schema.json",
   "name": "keboola-claude-kit",
-  "version": "1.9.0",
+  "version": "1.10.0",
   "metadata": {
     "description": "Comprehensive Claude Kit marketplace for Keboola workflows including developer tools, component development, and data app building with specialized agents and best practices"
   },
@@ -37,6 +37,13 @@
       "version": "1.0.0",
       "source": "./plugins/keboola-cli",
       "category": "operations"
+    },
+    {
+      "name": "sl-builder",
+      "description": "Generate, build, and populate semantic layer models for any Keboola project — works with or without the Semantic Layer data app",
+      "version": "1.0.0",
+      "source": "./plugins/sl-builder",
+      "category": "development"
     }
   ]
 }
--- a/plugins/sl-builder/.claude-plugin/plugin.json
+++ b/plugins/sl-builder/.claude-plugin/plugin.json
@@ -1,0 +1,9 @@
+{
+  "name": "sl-builder",
+  "version": "1.0.0",
+  "description": "Generate, build, and populate semantic layer models for any Keboola project — works with or without the Semantic Layer data app",
+  "author": {
+    "name": "Keboola s.r.o.",
+    "email": "support@keboola.com"
+  }
+}
--- a/plugins/sl-builder/skills/semantic-layer/SKILL.md
+++ b/plugins/sl-builder/skills/semantic-layer/SKILL.md
@@ -1,0 +1,542 @@
+---
+name: sl-builder
+description: >-
+  Use this skill to generate, build, or populate a semantic layer for any Keboola project.
+  Asks the user which project to use, fetches schemas and SQL transformations via kbagent,
+  generates a full semantic model (datasets, metrics, relationships, glossary), runs a
+  programmatic validation suite with auto-fixes, then pushes directly to the project metastore
+  after user confirmation. Supports partial update of existing models. No app dependency.
+  Use for: generate semantic layer, build semantic model, create metrics, define datasets,
+  sl-builder, semantic layer wizard, create data model, semantic layer from Keboola,
+  populate semantic layer, build me a semantic layer, analyze tables and create model.
+  Trigger any time the user mentions semantic layer, semantic model, or building metrics from data.
+---
+
+# sl-builder — Semantic Layer Builder
+
+Schema discovery → SQL context → business questions → generate → validate → confirm & push.
+
+---
+
+## Step 0 — Setup
+
+```bash
+export KBAGENT_CONVERSATION_ID="$(uuidgen)"
+python3 -c "
+import json,os
+cfg=json.load(open(os.path.expanduser('~/Library/Application Support/keboola-agent-cli/config.json')))
+for p in sorted(cfg.get('projects',{}).keys()): print(p)
+"
+```
+
+**Always ask the user to pick a project.** Never auto-select.
+
+Once chosen, extract credentials, list tables, and check existing models — all in parallel:
+
+```bash
+python3 -c "
+import json,re,os,urllib.request
+cfg=json.load(open(os.path.expanduser('~/Library/Application Support/keboola-agent-cli/config.json')))
+p=cfg['projects']['PROJECT']
+m=re.search(r'connection\.([\w-]+)\.gcp\.keboola\.com',p['stack_url'])
+region=m.group(1) if m else 'us-east4'
+token=p['token']
+metastore='https://metastore.'+region+'.gcp.keboola.com'
+print('TOKEN='+token)
+print('METASTORE='+metastore)
+req=urllib.request.Request(metastore+'/api/v1/repository/semantic-model',
+    headers={'X-StorageAPI-Token':token})
+existing=json.loads(urllib.request.urlopen(req,timeout=15).read()).get('data',[])
+print('EXISTING='+json.dumps([{'id':m['id'],'name':m.get('data',{}).get('name',m['name'])} for m in existing]))
+" &
+kbagent --json storage tables --project PROJECT > /tmp/tables.json &
+wait
+```
+
+Tables: `json.load(open('/tmp/tables.json'))['data']['tables']`
+
+If `EXISTING` has models, show their names and ask: **"Update one of these or create new?"**
+Store the chosen model ID as `UPDATE_ID` (or `None` for new).
+
+---
+
+## Step 1 — Business questions (before fetching schemas)
+
+> "Three quick questions:
+> 1. **What does this data model?**
+> 2. **Top KPIs the business tracks** — numbers on an exec dashboard
+> 3. **Which buckets/layers are gold/authoritative?** (or 'I don't know')"
+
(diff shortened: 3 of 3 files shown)


jordanrburger (changes requested):
Reviewed the full SKILL.md. Strong shape — the schema-discovery / SQL-context / validation / push pipeline is the right backbone, and §6 (downstream gotchas) is genuinely valuable institutional knowledge. A few things to address before merge.

## 🚨 Critical (block merge)

**`cleanup_orphans()` deletes every other semantic model's items in the project on a fresh create** (lines 372–379, called on line 425):

```python
def cleanup_orphans(type_path, keep_uuid):
    ...
    for i in all_items:
        if i.get('attributes',{}).get('modelUUID') != keep_uuid:
            delete(f"/api/v1/repository/{type_path}/{i['id']}")
```

On the CREATE path, `keep_uuid` is the brand-new model's UUID. Every existing item in the project has a different UUID by definition → all get deleted. Run `sl-builder` once in a project that already has another semantic model and you wipe that other model's datasets/metrics/relationships/glossary.

The Operational Hygiene §"Don't mass-delete via `cleanup_orphans` if the target project has multiple semantic models — verify the modelUUID filter actually filters before calling delete" already warns about this — but the code calls it unconditionally. The bare `except: pass` (lines 370/379) also hides any API failure during the fetch step, so the user thinks orphans got cleaned but the list call may have silently failed.

Recommend: guard with `if UPDATE_ID:` and remove the `cleanup_orphans` call on the create path entirely.

## 🟡 High

1. **GCP-only stack regex** (line 41) — `r'connection\.([\w-]+)\.gcp\.keboola\.com'`. Breaks on AWS-US/EU and Azure-EU stacks; fallback to `us-east4` won't match the user's actual region. Suggest `r'connection\.([\w-]+)\.(?:gcp|aws|azure)\.keboola\.com'` and detect the cloud (the metastore URL also needs `.aws.` / `.azure.`).

2. **Storage token printed to stdout in plaintext** (line 45) — `print('TOKEN='+token)`. Lands in Claude transcripts, terminal scrollback, telemetry. Suggest: write to a tmp file with `0600` perms and source from there, or stream into the next command via process substitution.

3. **"Partial update" is actually full-replace** (lines 410–412):
   ```python
   existing = {i.get('attributes',{}).get('name'): i['id'] for i in get_list(type_path, uuid)}
   new_names = {i[name_key] for i in items}   # ← computed, never used
   for name,eid in existing.items(): delete(...)   # ← deletes ALL existing
   ```
   PR body advertises *"Supports partial update of existing models"* but the implementation deletes everything then re-pushes. `new_names` is dead code. Either implement the diff or update the description.

4. **Constraints generated but never pushed** — Step 3 has a 30-line constraint section with severity-suffix encoding + Step 4 validates them, but `TYPES` in Step 5 (lines 399–404) omits `semantic-constraint`. The carefully-built `_critical/_warning/_healthy/_review` payloads land in `/tmp/model.json` and disappear. Add `(\"semantic-constraint\", model.get('constraints', []), \"name\")` to `TYPES`.

5. **`PROJECT` / `CONV_ID` are literal-string placeholders** (lines 40, 52, 86, 103, 127, 130). The skill never tells the agent to substitute them, and `'KBAGENT_CONVERSATION_ID':'CONV_ID'` sets the env var to the literal string `'CONV_ID'`, not the conversation ID exported in Step 0. Use shell vars consistently: export `PROJECT=$chosen_project`, then `os.environ['PROJECT']` inside heredocs.

6. **`UPDATE_ID` handoff is implicit** (line 397: `UPDATE_ID = None  # or the existing model UUID if updating`). Step 0 parses `EXISTING` and Step 5 branches on `UPDATE_ID`, but nothing wires them together. The agent has to spot this and replace `None` manually. Easy to miss → defaults to None → fresh-create path → triggers the critical bug above. Either export `UPDATE_ID` as an env var when the user picks an existing model in Step 0, or explicitly instruct \"replace `None` with the UUID from `EXISTING` if updating.\"

## 🟢 Medium

7. **SQL silently truncated at 3000 chars** (line 151) — `[:3000]` with no warning. Add a `len(combined) > 3000` log so the agent knows business logic is being clipped.

8. **`fqn()` hardcodes `\"KEBOOLA\"` as the database name** (lines 178, 382). Real Keboola projects use `KEBOOLA_<projectId>` (e.g., `KEBOOLA_293` in the metastore_experiment sample). Should pull from `kbagent storage info` or similar. As-is, every metric SQL will reference a non-existent database in most projects.

9. **`fqn()` defined twice** (lines 177, 381). Drift risk. Define once at the top, reference from both Step 3 and Step 5.

10. **Type auto-fix is inconsistent with the rule above it** — line 184: *\"Never override STRING to boolean/date\"*. Auto-fix accepts `NUMERIC → boolean` and `NUMERIC → string` as valid (line 249). Tighten the allow-list to match the documented rule.

11. **Time-dim auto-join is FI-app-specific** (lines 253–272) — column names `PERIOD`/`CODE_PERIOD_VALUE`/`PERIOD_CODE`/`DATE_KEY`, table prefixes `fact_`/`ft_`. Skill claims \"any Keboola project\" but only fires for FI-style models. Either generalize the column-name set or comment as \"FI heuristic; override `TIME_KEY_NAMES` for other domains.\"

12. **VERSION rule hardcodes literal strings `'Actual'`/`'Budget'`** (line 190). Fine for the FI app but silent NULL aggregates if a customer uses `'A'`/`'B'`, `'plan'`/`'actual'`, etc. At minimum: validate that the VERSION column's distinct values include `'Actual'` and `'Budget'` before applying the rule.

13. **`kbagent` dependency unstated** — every shell call assumes `kbagent` installed; README/plugin.json don't list it. Customer environments may have `kbc` only. Add a preflight check (`command -v kbagent || ...`).

14. **No tests, no jsonschema validation** — verification is entirely \"agent re-runs Step 4 until ✓ clean,\" which puts the validation burden on the agent. Recommend adding a `tests/test_smoke.py` similar to #72 that round-trips a synthetic model through `jsonschema.validate` against `semantic-*_schema_1.0.0.json`. Catches regressions like the `sql_dialect` vs `sqlDialect` divergence automatically.

## ⚪ Nits (no need to fix, just noting)

- Shadowed `m` in Step 0 list comp (line 50) — readable but lints will flag.
- Step 6.5 promotion example references `fetch()` / `post_to_target()` that aren't defined. Pseudocode is fine; add `# pseudo-code:` so copy-pasters don't hit NameError.
- Frontmatter `description` is 9 lines / 10+ \"Use for:\" trigger phrases. Works for discoverability but heavier than typical Anthropic skill style. Style nit.

## Marketplace coordination

This PR + #72 (powerbi-to-sl) both bump `marketplace.json` minor version — whoever lands second rebases trivially (keep both plugin entries + bump version once more). #70 is currently DIRTY for unrelated reasons; rebase + the fixes above and it's mergeable.

ottomansky:
Closing in favour of a new PR that upgrades `sl-builder` from v1 to v2.

**Why:** The original PR shipped a single monolithic build-wizard skill. After live testing against a real Keboola project, the plugin was split into 6 focused commands (`sl-show`, `sl-validate`, `sl-add`, `sl-edit`, `sl-remove`, `sl-build`) — each independently invocable. `sl-edit` also gained automatic cascade updates for constraint references on metric rename, which the monolith didn't handle. The new PR replaces this one entirely.

PR #71, opened 2026-04-29
feat (dataapp-deployment): add knowledge about Storage access 

MiroCillik:
## Why

Developers building Keboola Data Apps that need live Storage I/O (reads and writes through the Query Service) had no skill-level guidance. Without it, an AI-assisted developer was on their own to discover the `direct-grant` output mapping, the four platform-injected env vars, the exact table-identifier shape Snowflake expects, and — most dangerously — the fact that the Query Service has no parameterized queries, which makes every interpolated value a potential injection point.

The result: the skill produced container-correct apps that didn't actually know how to talk to Storage, so the developer had to leave the skill to figure out the runtime layer.

## Context

The existing `dataapp-deployment` skill stops at the container boundary — it covers Nginx, Supervisord, secrets, and the POST-`/`-on-startup quirk, but says nothing about how an app reads or writes a Storage table. Meanwhile, interactive read/write data apps (dashboards with input forms, lightweight admin tools, internal one-pagers) are increasingly the primary use case, and Keboola already ships a Python and a JS Query Service SDK that make this practical. The gap was knowledge, not capability.

## Approach

Document the full runtime layer as a single section that takes a developer from operator-side configuration through to "first INSERT works locally." The narrative is:

1. **Operator side** — what the Storage Access toggle does, what `unload_strategy: "direct-grant"` actually grants, and that bucket stage (`in.` vs `out.`) does not constrain writes (a misconception that bit us during testing).
2. **Runtime side** — the four env vars the platform injects, copy-paste wrappers for both Python (`keboola-query-service`) and Node.js / TypeScript (`@keboola/query-service`), and the canonical table-identifier syntax.
3. **Security** — every SQL value must be validated and escaped because the Query Service does not support parameterized queries. Both languages get a concrete validation module example.
4. **Local development** — the workspace + token + `.env` workflow so the same code runs identically locally and in the container.
5. **Errors** — the four failure modes a developer is actually going to hit, with the fix for each.

Python and Node.js / TypeScript are covered side-by-side throughout because the skill targets both runtimes — the container image ships both — and an AI assistant choosing between them shouldn't have to discover the JS SDK separately. The Streamlit and Express subsections each get a small "storage access in this framework" pointer (Streamlit: `@st.cache_resource` for the singleton; Express: import-time env-var requirement) so the per-framework patterns close the loop with the new section.

## Impact

- An AI-assisted developer can scaffold a working Data App with live Storage I/O on first try, in either Python or JavaScript, without leaving the skill.
- The validation guidance closes a real injection-shaped trap that the previous skill silently left open.
- The frontmatter description now triggers the skill for Streamlit / FastAPI / Flask and Node.js / TypeScript prompts, not just generic "deploy a data app" prompts.
- No breaking changes; this is purely additive.

## Testing

Validated by building two greenfield apps directly from the new skill content before opening this PR:

- Python + FastAPI app reading and writing `out.c-data-app.mvc-crashes` (Chart.js + Leaflet visualization, add-crash form).
- Node.js + Express app reading and writing `in.c-dbt-beers.beers` (table browser with filter, add-beer form).

Both apps ran end-to-end against project 40 on `eu-central-1` on first try — KPIs populated, the map and table rendered, and INSERTs round-tripped. The `in.c-` write target on the second app is what surfaced the bucket-stage clarification now in the doc.

## Notes

- Source-of-truth file is `plugins/dataapp-developer/skills/dataapp-deployment/SKILL.md`. Anyone using a personal copy at `~/.claude/skills/keboola-data-app/SKILL.md` will need to resync after merge.
- The two reference apps (`data-app-direct-access-skill`, `data-app-direct-access-js`) are not part of this repo and don't need to land here, but they're available on request as worked examples for future updates.

### DIFF
--- a/plugins/dataapp-developer/skills/dataapp-deployment/SKILL.md
+++ b/plugins/dataapp-developer/skills/dataapp-deployment/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: dataapp-deployment
-description: Use when deploying any web app to Keboola Data Apps, setting up keboola-config directory, configuring Nginx/Supervisord for Docker, handling SSE or WebSocket streaming through Nginx, mapping secrets to environment variables, or debugging Keboola Data App deployment issues like POST to root errors, 500s from missing env vars, or buffered streams.
+description: Use when deploying any web app (Python — Streamlit/FastAPI/Flask — or Node.js / TypeScript) to Keboola Data Apps, setting up keboola-config directory, configuring Nginx/Supervisord for Docker, handling SSE or WebSocket streaming through Nginx, mapping secrets to environment variables, accessing Keboola Storage via the Query Service (direct-grant output mapping, keboola-query-service Python or @keboola/query-service JS client), local development setup with workspace tokens, or debugging Keboola Data App deployment issues like POST to root errors, 500s from missing env vars, or buffered streams.
 ---
 
 # Deploying Web Apps to Keboola Data Apps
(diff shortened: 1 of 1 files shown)


jordanrburger:
@MiroCillik - Maybe we can sit down together for 30 minutes and get the skill completely updated. I don't think the current structure is good at all. I've been meaning to change it but I haven't gotten around to it. Imho, we need to unify everything under "dataapp-developer" and get rid of the data app deployment skill all together.

MiroCillik:
@jordanrburger yes Jordan, I completely agree, would be best to unify those together.

davidesner:
> @MiroCillik - Maybe we can sit down together for 30 minutes and get the skill completely updated. I don't think the current structure is good at all. I've been meaning to change it but I haven't gotten around to it. Imho, we need to unify everything under "dataapp-developer" and get rid of the data app deployment skill all together.

@MiroCillik FYI https://linear.app/keboola/issue/AI-3147/extend-data-app-development-skill-to-cover-full-lifecycle-storage-git

MiroCillik:
I'm closing this, David Esner is preparing a new consolidated skill for Data Apps including Direct Storage access

MiroCillik:
Superseded by https://github.com/keboola/ai-kit/pull/76

PR #72, opened 2026-05-13
feat(powerbi-to-sl): Power BI → Keboola SL migration plugin

jordanrburger:
## Summary

Adds the **`powerbi-to-sl`** plugin: an end-to-end skill that translates an existing Microsoft Power BI semantic model into a Keboola semantic layer model conforming to the metastore service schemas.

Brownfield companion to [`sl-builder`](https://github.com/keboola/ai-kit/pull/70). Where `sl-builder` generates a new SL from Keboola tables + business questions (LLM-driven), this skill performs **mechanical structural translation** of an existing Power BI model — tables, columns, measures, relationships — and hands the resulting payloads to `sl-builder`'s push pipeline (or to a direct REST snippet).

## Why two skills, not one

| Aspect       | `sl-builder`                          | `powerbi-to-sl` (this)            |
|--------------|---------------------------------------|-----------------------------------|
| Direction    | Greenfield (Keboola tables → SL)      | Brownfield (Power BI → SL)        |
| Input        | Keboola storage + SQL transformations | TMDL folder or per-table JSON     |
| Generation   | LLM-driven from schema + KPIs         | Mechanical structural translation |
| Validation   | Auto-fix + LLM critique               | jsonschema validate + WARNINGS    |
| Push         | Owns push pipeline                    | Hands off to `sl-builder` push    |

Reuse, not overlap.

## Inputs

Two shapes are accepted (auto-detected, override via `--input-format`):

1. **TMDL folder** — extracted via the Microsoft Power BI Modeling MCP server in `--readonly` mode. Canonical input — captures DAX **and** Power Query M expressions verbatim.
2. **Per-table JSON** — one `*_semantic_layer.json` file per Power BI table, produced by older `get_semantic_model` calls. Missing M steps; acceptable for structural migration when full fidelity isn't required.

## Output

JSON files matching the Keboola metastore `semantic-*_schema_1.0.0.json` schemas:

```
out/
  semantic-model.json
  semantic-dataset/<table-slug>.json
  semantic-metric/<measure-slug>.json    # DAX preserved verbatim
  semantic-relationship/<from>__<to>.json
  WARNINGS.md                            # complex DAX, unmapped types, dangling refs
```

## Mapping highlights

- `tables[]` → `semantic-dataset` (slugified `tableId` from configurable bucket prefix)
- `columns[].dataType` → `fields[].type` (`int64`→`integer`, `double`/`currency`→`decimal`, `dateTime`→`datetime`, etc.)
[...]

PR #74, opened 2026-05-14
feat(sl-toolkit): v2 — 3 commands, CRUD conversational, multi-cloud, cascade rename (renamed from sl-builder)

ottomansky:
Replaces #70 (closed).

## Architecture

v1 shipped a monolithic build-wizard skill. After live testing and reviewer feedback, v2 is slimmed to **3 explicit slash commands** — operations that users naturally type as commands. Add / edit / remove are handled **conversationally** via the `semantic-layer` reference skill (users say "add a metric for Revenue", not `/sl-add`).

| Command | What it does |
|---|---|
| `/sl-show` | List all datasets, metrics, relationships, constraints, glossary |
| `/sl-validate` | Check for phantom fields, dangling refs, constraint orphans, AGG-on-STRING; `--deep` hits Snowflake via kbagent |
| `/sl-build` | Greenfield wizard: schema discovery → AI generation → validate → push |

`/sl-add`, `/sl-edit`, `/sl-remove` are deleted as commands. Their procedures (with rollback and cascade rename) now live inline in `SKILL.md`.

## Resolves Jordan's review (#74)

All 14 items from Jordan's `CHANGES_REQUESTED` review are addressed. Status table:

| # | Severity | Item | Resolution |
|---|---|---|---|
| C1 | 🚨 Critical | `/sl-build` push loop missing `semantic-constraint` | ✅ Fixed in earlier commits — `commands/sl-build.md` push loop includes `semantic-constraint` |
| H1 | 🟡 High | GCP-only stack regex | ✅ Fixed in earlier commits — `(gcp\|aws\|azure)` alternation in `SKILL.md` |
| H2 | 🟡 High | Literal placeholders (`TOKEN`, `METASTORE`, `CONV_ID`, `PROJECT_ALIAS`) | ✅ Fixed in earlier commits; verified clean by `test_no_literal_placeholder_strings_in_python_heredocs` |
| H3 | 🟡 High | `UPDATE_ID` handoff between Python blocks | ✅ Fixed in earlier commits — persisted to `/tmp/sl_update_id.txt` |
| H4 | 🟡 High | `sl-edit` no rollback after DELETE | ✅ Fixed in earlier commits — `original_attrs` saved, re-POST on failure (`SKILL.md` Edit procedure) |
| M5 | 🟢 Medium | Unescaped `.` in deep-check regex | ✅ Fixed in earlier commits — `\.` in both `re.findall` calls |
| **M6** | 🟢 Medium | **`fqn()` duplicated + hardcoded `KEBOOLA` DB** | ✅ **Fixed in this push** — new `db_name()` helper in `SKILL.md` API Primitives queries `/v2/storage/tokens/verify` and caches `KEBOOLA_{projectId}` (e.g. `KEBOOLA_293`) to `/tmp/sl_db_name.txt`. Both `fqn()` call sites use it. Fallback to bare `KEBOOLA` with stderr warning on resolve failure. |
| **M7** | 🟢 Medium | **VERSION rule hardcodes `'Actual'`/`'Budget'`** | ✅ **Fixed in this push** — new `/sl-build` **Step 2.5** probes distinct values for `VERSION/SCENARIO/PERIOD_TYPE/TYPE/VARIANT` columns via parallel `kbagent` SQL (10 workers). Step 3 only emits VERSION-conditional metrics if values match the `{actual, budget, plan, forecast, baseline, target}` family. `SKILL.md` VERSION rule rewritten to instruct probing. |
| M8 | 🟢 Medium | `sl-edit` cascade metric-only — needs scope warning | ✅ Fixed in earlier commits — explicit `⚠ Dataset/relationship renames are not cascaded` note in `SKILL.md` |
| M9 | 🟢 Medium | `sl-validate --deep` sequential | ✅ Fixed in earlier commits — `ThreadPoolExecutor(max_workers=10)` |
| M10 | 🟢 Medium | Step 4 validation source — local file vs metastore API | ✅ Fixed in earlier commits — reads `/tmp/sl_model.json` + `/tmp/sl_schemas.json` |
| **L11** | ⚪ Low | **`api_get/api_post/api_delete` defined but unused** | ✅ **Fixed in this push** — `SKILL.md` API Primitives are now the canonical HTTP layer. CRUD blocks in `SKILL.md` (Add/Edit/Remove) and the three command files call `api_*` instead of inline `urllib`. Each command's first Python block defines local copies (heredocs run as fresh processes). |
| L12 | ⚪ Low | `allowed-tools: ['*']` on reference skill | ✅ Fixed in earlier commits — `allowed-tools` removed from `SKILL.md` frontmatter |
| L13 | ⚪ Low | Trigger overlap with `/sl-add`/`/sl-edit` | ✅ Fixed in earlier commits — commands deleted in v2; triggers cleaned up |
| **L14** | ⚪ Low | **No tests / no jsonschema validation** | ✅ **Fixed in this push** — 30-test suite in `plugins/sl-builder/tests/` covering jsonschema round-trip for all 6 entity types, `db_name()`/`fqn()` unit tests, and SKILL.md consistency greps |

### Bonus regression caught by the new tests

The new `test_skill_consistency.py::test_sqldialect_is_camelcase` and `test_smoke.py::test_no_dialect_drift` caught a real regression while being written: `SKILL.md:138` (the `semantic-model` POST envelope example) and `commands/sl-build.md:302` (model creation POST) both still used snake_case `sql_dialect` — **the exact bug Jordan caught in PR #72**, lingering in two un-updated call sites. Both are now `sqlDialect`. The test net is what Jordan asked for.

## Bug fixes from earlier code review (pre-#74)

**Critical**
- `sl-build` Step 5 push loop was missing `semantic-constraint` — constraints were never pushed to the metastore

**High**
- Multi-cloud regex: `gcp`-only pattern now matches `gcp|aws|azure` stacks in both auth fallbacks
- `sl-build` literal placeholders (`token='TOKEN'`, `metastore='METASTORE'`, `'CONV_ID'`) replaced with resolved values from setup recipe
- `UPDATE_ID` now persisted across Python blocks via `/tmp/sl_update_id.txt`
- `sl-edit` now saves `original_attrs` before DELETE and attempts re-POST on POST failure (rollback)

**Medium**
- `sl-validate` deep-check regex: unescaped `.` fixed to `\.` in both `findall` calls
- `sl-validate` deep check parallelized with `ThreadPoolExecutor(max_workers=10)`
- `sl-build` Step 4 validation explicitly reads local `/tmp/sl_model.json` — not the metastore API

## Other changes
- `semantic-layer` SKILL.md: removed `allowed-tools: ['*']`, trimmed auto-load triggers, updated command list, added full CRUD Operations section with cascade rename + rollback + constraint-orphan check
- `marketplace.json`: merge conflict resolved; kit bumped to v1.10.0
- `plugin.json` and READMEs updated to reflect 3-command structure

## Test plan

- [x] `python -m pytest plugins/sl-builder/tests/` — 30/30 pass locally
- [ ] `/sl-show` — lists model entities grouped by type with correct counts
- [ ] `/sl-validate` — catches phantom fields, constraint orphans; `--deep` checks Snowflake schema
- [ ] `/sl-build` — full greenfield wizard; verify constraints are pushed alongside other entity types
- [ ] `/sl-build` — verify `/tmp/sl_db_name.txt` is created with `KEBOOLA_{projectId}` and generated metric SQL uses it
- [ ] `/sl-build` Step 2.5 — verify VERSION-candidate columns are probed and metrics skip-with-warning when distinct values don't match the expected family
- [ ] Conversational add/edit/remove via `semantic-layer` skill — confirm CRUD works without slash commands
- [ ] Edit → rename a metric; verify constraints auto-cascade; verify rollback fires if POST fails
- [ ] Run on a non-GCP stack (AWS-EU or Azure) to confirm multi-cloud regex resolves correctly


jordanrburger (changes requested):
Big architectural improvement over #70 — the multi-command split is the right call, the critical `cleanup_orphans` destructive-delete bug is gone, the kbagent-free CRUD path is genuinely useful, and the constraint cascade on rename is a thoughtful safety feature. Tracking carryovers from the previous review + a few new things below.

## ✅ Fixed from #70

- 🚨 **`cleanup_orphans` deleting other models' items** — removed entirely. The most important fix.
- 🟡 Storage token printed to stdout — gone from all setup paths.
- 🟡 "Partial update is actually full-replace" — the workflow changed; per-item commands handle this cleanly now.
- 🟢 SQL silent truncation at 3000 chars — extraction block removed.
- 🟢 Type auto-fix inconsistent with documented rule — auto-fix removed, now a warning.
- 🟢 FI-specific time-dim auto-join — removed.
- 🟢 kbagent dependency unstated — explicit fallback chain + `command -v kbagent` detection.

## 🚨 Critical (still — block merge)

**Constraints generated but never pushed in `/sl-build`** — same bug as #70. `commands/sl-build.md` lines 159–164:

```python
for type_path, items, key in [
    ('semantic-dataset', model['datasets'], 'name'),
    ('semantic-metric',  model['metrics'],  'name'),
    ('semantic-relationship', model['relationships'], 'name'),
    ('semantic-glossary', model['glossary'], 'term'),
]:
```

`semantic-constraint` still missing from the push loop. Step 3 says to use the constraint encoding from SKILL.md, and SKILL.md has the whole 4-band severity-suffix spec, but greenfield builds never persist them. Add `('semantic-constraint', model.get('constraints', []), 'name')` to the tuple list. Step 3's JSON-keys hint (line 106) also needs `constraints` added so the generator builds them.

## 🟡 High

1. **GCP-only stack regex** — still hardcoded `r'connection\.([\w-]+)\.gcp\.keboola\.com'` in `SKILL.md` lines 31 and 43. Two duplicated copies, both GCP-only. AWS-US/EU and Azure-EU stacks will hit the `us-east4` fallback and POST to the wrong metastore (silent — POST will 404 or worse, succeed against an unintended region). Suggest:
   ```python
   m = re.search(r'connection\.([\w-]+)\.(?:gcp|aws|azure)\.keboola\.com', stack_url)
   cloud = re.search(r'\.(gcp|aws|azure)\.keboola\.com', stack_url).group(1) if stack_url else 'gcp'
   METASTORE = f'https://metastore.{m.group(1)}.{cloud}.keboola.com'
   ```

2. **`PROJECT_ALIAS` / `CONV_ID` literal placeholders carry over.** Specific spots:
   - `SKILL.md:42` — `p = cfg['projects']['PROJECT_ALIAS']`
   - `sl-build.md:35–36` — `token='TOKEN'; metastore='METASTORE'` (these are literal strings, not vars)
   - `sl-build.md:79` — `ENV = {**os.environ, 'KBAGENT_CONVERSATION_ID': 'CONV_ID'}` overwrites the exported env var with the literal string `'CONV_ID'`
   - `sl-build.md:82,97` — `--project PROJECT`
   - `sl-validate.md:103` — `--project','PROJECT_ALIAS'`

   For agent-driven skills the LLM is meant to substitute, but mixing `$PROJECT` shell refs and literal `'PROJECT'` Python strings in adjacent blocks confuses the substitution model. Pick one: bash-exported `$PROJECT` and read via `os.environ['PROJECT']` inside heredocs, or jinja-style `{{PROJECT}}` markers everywhere. Currently the agent has to guess which spots to rewrite.

3. **`UPDATE_ID` handoff still implicit** — `sl-build.md` Step 0 says "Store chosen model id as `UPDATE_ID` (or `None` for new)" without showing how to persist it. Step 5 line 149 reads `UPDATE_ID`. Across two separate Python blocks in different bash invocations, the variable doesn't survive. Either export to env in Step 0 and read in Step 5, or write to `/tmp/sl_update_id` and read back.

4. **`sl-edit` has no rollback if POST fails after DELETE** — lines 84–100. DELETE removes the old metric, then POST creates the new one. If POST fails (HTTP 5xx, validation error, network), the old metric is gone with no replacement. The previous version's `push()` had retry-3-with-sleep; that's gone. Suggest: keep the pre-DELETE payload in memory; on POST failure, attempt to re-POST the original.

## 🟢 Medium

5. **Regex regression in `sl-validate.md` lines 117, 120** — the dot between table and column is unescaped:
   ```python
   for col in re.findall(r'"[^"]+"."([^"]+)"', m.get('sql',''))
   for col in re.findall(r'\b(?:SUM|AVG)\s*\(\s*"[^"]+"."([^"]+)"\s*\)', m.get('sql',''), re.I)
   ```
   `.` matches any character. PR #70 had `\.` (escaped). Tighten to `r'"[^"]+"\."([^"]+)"'` to avoid spurious matches on SQL like `"FACT"x"AMOUNT"`.

6. **`fqn()` defined twice + hardcodes `"KEBOOLA"`** — `SKILL.md:140–142` and `sl-build.md:137–138`. Real Keboola projects use `KEBOOLA_<projectId>` (the `metastore_experiment` sample shows `KEBOOLA_293`). Every metric SQL will reference a non-existent DB in most projects. Centralize and pull the DB name from `kbagent storage info` or a credentials probe.

7. **VERSION rule hardcodes `'Actual'` / `'Budget'` literals** — `SKILL.md:167`. Silent NULL aggregates if a customer uses `'A'/'B'` or `'plan'/'actual'`. At minimum validate that the column's distinct values include the expected strings before applying the rule.

8. **`sl-edit` cascade is metric-only** — datasets and relationships also have downstream consumers. Renaming a dataset's `tableId` invalidates every metric's `dataset` field and every relationship's `from`/`to`. Out of scope to fix here, but worth a `⚠ scope` note in the command body so users know not to expect cascade for dataset renames.

9. **`sl-validate --deep` table-detail fetch is sequential** — `sl-validate.md:101–123` runs one `kbagent` subprocess per dataset in a for-loop. `sl-build.md:87–91` parallelizes the same call with 10 workers. For a 101-metric / 30-table model, sequential is minutes vs. seconds. Use `ThreadPoolExecutor` here too.

10. **`/sl-build` Step 4 says "Run the checks from `/sl-validate` inline"** — but `/sl-validate` reads from the metastore API (`api_get`), while `/sl-build`'s model lives at `/tmp/sl_model.json` (not yet pushed). The validation logic needs to source from the local file in the build context. Either factor the checks into a function that takes `(datasets, metrics, ...)` lists, or duplicate the rule block in `sl-build.md` operating on `json.load(open('/tmp/sl_model.json'))`.

## ⚪ Low / nits

11. **`api_get`/`api_post`/`api_delete` primitives defined in `SKILL.md:65–79` but never used** — every command re-implements urllib inline. Either commands should call the primitives (reduces duplication and makes future API changes one-line), or delete the primitives from the skill (they're dead documentation).

12. **`allowed-tools: ['*']` on a reference skill** — `SKILL.md:11`. Reference skills don't execute, they just provide context — `allowed-tools` is moot here. The commands correctly scope to `Bash` + `AskUserQuestion`. Either drop the field or scope it.

13. **Triggers in `SKILL.md:4–8` include "add metric", "edit metric", etc.** — these compete with `/sl-add` and `/sl-edit` commands. Both fire? Just the skill? Anthropic's pattern is usually: commands are user-typed, skills auto-load on context match. Either drop the verb triggers from the skill description (let commands handle imperative phrasings) or document the layered behavior in the plugin README.

14. **No tests / no jsonschema validation** — carryover from #70. The `sql_dialect` vs `sqlDialect` divergence I caught in #72 is exactly the class of bug a `tests/test_smoke.py` with `jsonschema.validate` would surface automatically. The new code paths (POST envelopes for each entity type) have more surface area for this to recur. Recommend at least a fixture-based round-trip test for each entity type.

## Marketplace coordination

Branch is `feat/sl-builder-v2`; rebased on top of #72 (merged), so my `powerbi-to-sl` entry is preserved in `marketplace.json`. Version bump from 1.9.0 → 1.10.0 is clean. PR is currently `DIRTY` per `gh` — likely a small main-drift; re-rebase should clear it.

---

**Recommendation:** Address the critical (constraint push) and the high-severity carryovers (regex, placeholders, UPDATE_ID handoff, sl-edit rollback). Medium and nit items can ride in a follow-up. Strong improvement over #70 — happy to re-review after the next push.

ottomansky:
Renamed `sl-builder` → `sl-toolkit` per your point — v2 is no longer just a greenfield builder, it's the lifecycle kit (build / show / validate + conversational CRUD). `sl-toolkit` describes the scope honestly and matches the parent `ai-kit` naming.

**What changed**
- `plugins/sl-builder/` → `plugins/sl-toolkit/` (single `git mv`, git rename detection clean — all 25 files preserved with history)
- `plugin.json`, `marketplace.json`: name + source updated
- Plugin version: 2.0.0 → 3.0.0 (slug change is breaking for installers)
- Kit version: 1.10.0 → 1.11.0
- Cross-references in `powerbi-to-sl` (plugin.json + README + SKILL.md, 14 hits) updated
- Top-level + nested READMEs updated
- Migration breadcrumbs in `plugins/sl-toolkit/README.md` changelog and the marketplace description so anyone moving from the old slug can find it

**What did not change**
- Commands: `/sl-build`, `/sl-show`, `/sl-validate` — keep names (the `sl-` prefix already namespaces them)
- Reference skill: still named `semantic-layer` (it's domain-named, not plugin-named)
- All test fixtures + schemas — zero references to the old slug

**Verification**
- `grep -rn sl-builder` only returns intentional migration breadcrumbs (3 in the changelog + 1 in marketplace description)
- All 30 tests pass under the new path: `python -m pytest plugins/sl-toolkit/tests/ -v` → 30 passed in 0.07s
- JSON files parse cleanly

Pushed to `feat/sl-builder-v2` (commit `5182e5e`). Ready for re-review.

jordanrburger (approved):

PR #76, opened 2026-05-18
Consolidate Keboola App skills into dataapp-development

davidesner:
## Summary

Consolidates two legacy skills (`dataapp-dev` for Streamlit, `dataapp-deployment` for Python/JS) into a single new `dataapp-development` skill at `plugins/dataapp-developer/skills/dataapp-development/`. The new skill is a router (`SKILL.md` ≈ 100 lines + 13 references + 5 runnable templates) that covers the full app lifecycle across both app types and all three client paths.

### What's covered

- **App types:** Streamlit, single-Node + static (the dashboarding default), combined Python+Node
- **Client paths:** MCP-only (Claude Desktop / web), Claude Code with filesystem + MCP, kbagent CLI — with sequential detection so an agent always offers every available path instead of silently picking
- **Storage access:** read-only workspace + Query Service SDKs (`keboola-query-service` / `@keboola/query-service`), DuckDB caching as default for read-only apps, RW Storage Access for writes, input mapping discouraged. BigQuery's `workspace_query` path documented separately
- **Local dev:** proactive `.env.local` pre-fill from `get_project_info`, ask user only for `KBC_TOKEN`, with explicit guidance that narrow-scoped tokens don't work with Query Service
- **Styling:** default Keboola palette across all three stacks, "Powered by Keboola" footer with brand-override removal instructions, heavier React+Vite+shadcn option for complex UIs
- **Other:** authentication, dashboard patterns, optional Kai chat integration, troubleshooting (10+ live-test-driven entries), dev workflow

### Templates

- `templates/streamlit/` — keboola-query-service SDK + Plotly + footer
- `templates/nodejs-app/` — Express + Tailwind CDN + Chart.js + footer (dashboarding default)
- `templates/python-app/` — Flask + uv
- `templates/python-node-app/` — FastAPI backend + Express frontend
- `templates/duckdb-cache/` — Python and Node DuckDB cache helpers

### Hard rules

9 SKILL.md hard rules that emerged from live test sessions:
1. Never commit secrets
2. RO workspace before input mapping
3. Apps must handle POST /
4. No `pip install` (PEP 668 — use `uv`)
5. No `[program:nginx]` declaration
6. Validate data first, code second (semantic layer check)
7. Pick one Keboola path per session (sequential detection, list all candidates)
8. MCP-only flows: compose source directly into the tool call — don't pre-write a local copy
9. For local-dev credentials: pre-fill what you can, ask only for what's missing, never grep the filesystem

### Validated against

Three end-to-end test sessions against a real Keboola project. Each surfaced a specific footgun that's now codified in the skill (Path A doubled emit, kbagent omission from path detection, legacy workspace-query 404, narrow-scoped token auth error, agent scanning filesystem for tokens).

### Removed

- `plugins/dataapp-developer/skills/dataapp-dev/`
- `plugins/dataapp-developer/skills/dataapp-deployment/`

## Test plan

- [ ] Spot-check `SKILL.md` decision tree routes to the right reference for each task type
- [ ] Verify the three templates that ship a "Powered by Keboola" footer render correctly (nodejs-app, python-node-app, streamlit)
- [ ] One more end-to-end run from a fresh Claude Code session against a real project: build a new Streamlit app, deploy, debug
- [ ] One more end-to-end run for a Python/JS app via the kbagent path
- [ ] Plugin version + marketplace version bumped and root README feature list updated


ottomansky:
Hey @davidesner  summarizing my findings 
Overall this is really solid — I got a working data app running locally in ~15–20 minutes end-to-end, which is genuinely impressive for a first pass. I didn't push it into Keboola, just ran it locally, so this is purely feedback from the build path. A handful of rough edges worth flagging:

**Stack selection (App type → Project step)**
- Would be nice if the list sorted by most-used stacks first — I started not sure which one I needed.
- "AWS US (default)" says `connection.keboola.com`, but the bundled MCP URL is `mcp.us-east4.gcp.keboola.com` — a GCP hostname under the AWS US option, which threw me off (is this AWS or GCP?). Worth either tweaking the labeling or adding a note that the MCP runtime lives on GCP regardless of the storage stack.

**Bundled MCP URL is US-only**
- `plugin.json` hardcodes `mcp.us-east4.gcp.keboola.com`. I'm on GCP EU (europe-west3), so it returned nothing and I fell back to kbagent. Either make it stack-aware (derive from the user's project URL) or call it out in the README so non-US users know to expect the kbagent path.

**`storage-access.md` assumes newer MCP fields**
- It says to call `get_project_info` and read `workspace_id` / `branch_id`, but on `keboola-mcp-server` v1.32.0 those fields aren't returned — only `project_id`, `project_name`, `sql_dialect`, etc. I had to detour through `kbagent branch list` + `kbagent workspace list` to find them. Either bump the minimum MCP version in the skill or document the fallback.

**kbagent `--json` is fragile**
- Every kbagent call prints `Updating keboola-mcp-server v1.32.0 -> v1.61.3 (via uv_tool)...` to stdout, which breaks `json.loads`. I had to regex-strip the leading lines before every parse. The upgrade nag should go to stderr or sit behind a `--quiet` flag, otherwise scripting against `--json` is painful. -> I will create an issue in https://github.com/padak/keboola_agent_cli and try to create a PR for fixing this or discussing with Padak

**kbagent project setup isn't mentioned**
- When I picked the kbagent path, `kbagent project list` showed 3 unrelated projects, not the one I was working on. The skill never mentioned I might need `kbagent project add` first. The path-selection step could say: "if you chose kbagent, confirm `kbagent project list` includes your project; if not, run `kbagent project add` first."

**FQN fallback in `api/queries.js`**
- The template comment tells you to copy the FQN from `mcp__keboola__get_table.fully_qualified_name`, but on the older MCP that field doesn't exist (table-detail returned `fully_qualified_name: None`). I ended up constructing it manually as `"<DATABASE>"."<BUCKET>"."<TABLE>"` using the database name from `kbagent workspace list`. The comment should at least mention the manual construction rule as a fallback.

**Playwright version pin**
- `npx playwright install` defaults to latest, but the MCP server pins `playwright@1.57.0` (Chromium build v1200), so the first two screenshot attempts failed. Took an `npm view @executeautomation/playwright-mcp-server dependencies` to figure out the right version. Could either pin it in the setup instructions or auto-detect.

**Token-in-chat ergonomics**
- One of the AskUserQuestion options for the token was effectively "paste it here" — and I did, which means it lives in the transcript even after I rotate. Would be cleaner if the agent always pointed at `.env` first and treated "paste in chat" as the discouraged escape hatch, with a stronger warning. 😬

**`kbagent project add --url` default**
- Defaults to `https://connection.keboola.com` (legacy AWS US stack), which is wrong for most projects today. Should default to empty/required, or auto-detect from a provided project URL.

None of these blocked me from getting to a working app, but smoothing them out would make the first-run experience a lot tighter — especially for non-US-AWS users. Happy to help test a follow-up pass.

ottomansky:
Also deployed into keboola overall took 3-5 minutes from the time I asked it on the locally running app CC pointed this out 
  What this run added to the PR feedback bucket

  Three meaningful gaps caught:
  
  1. templates/nodejs-app/ is not deployable as shipped — validate-repo fails with 3 BLOCKING issues. Missing keboola-config/nginx/sites/default.conf, keboola-config/supervisord/services/app.conf, and pyproject.toml. The python-app and python-node-app
  templates have these; the Node template should too. I copied + adapted from python-app.
  2. pyproject.toml is required even for pure-Node apps — validate-repo --type python-js enforces it. Either the validator should skip this check when no Python is detected, or the docs should call out the stub-pyproject pattern explicitly.
  3. data-app password ergonomics with manage-token blocking — the security model (default-deny KBC_MANAGE_API_TOKEN) is good, but the failure message "No manage token available. Run interactively, or pass --allow-env-manage-token" surfaces only after
  the call, not as a hint when kbagent data-app create --auth password was chosen. Could nudge earlier: "you picked password auth — to fetch the password later, you'll need --allow-env-manage-token or the UI."

ottomansky:
One more finding is that kbagent currently doesnt have a way to look into terminal logs so it cant debug the apps when there is some issue starting them and gets into loop, already working on a PR for that cc: @jordanrburger

AdamVyborny (approved):
Merging as is, we are already using it. Let's address any issue is follow ups

PR #81, opened 2026-06-17
fix: align dataapp-developer skill with renamed modify_streamlit_data_app MCP tool

sykora-ji:
[link to issue](https://linear.app/keboola/issue/AJDA-2852/fix-mcp-for-streamlit-and-cc)

## Description

The Keboola MCP server renamed the Streamlit data-app tool `modify_data_app` → `modify_streamlit_data_app` (`keboola/mcp-server` AI-3005, 2026-05-24) and split out a separate `modify_python_js_data_app`. The `dataapp-developer` skill still instructed agents to call the old `modify_data_app`, which no longer exists on the server.

An agent following the skill calls `modify_data_app`, the tool is not found, and the call fails — with no working fallback, since `update_config` is intentionally blocked for data-app components by `check_suitable`.

This change renames all 24 references from `modify_data_app` to `modify_streamlit_data_app` across the skill (`SKILL.md`, `TODO.md`, and the `references/` files: `authentication.md`, `choosing-app-type.md`, `deployment-paths.md`, `glossary.md`, `storage-access.md`, `streamlit-apps.md`). It is a pure rename — the tool's parameters (`source_code`, `change_description`, `configuration_id`) are unchanged, so the documented call signatures stay accurate.

## Release Notes

- **Justification**
    - The MCP side of AJDA-2852 point 1 (two tool names for one operation, only one works) is already resolved on the server by the rename; the remaining gap was the agent guidance, which still pointed at the dead tool name. This aligns the `dataapp-developer` skill with the live MCP tool surface.
    - In plain terms: the assistant's instructions told it to use a tool that had been renamed, so it would fail when trying to create or edit a Streamlit data app. The instructions now use the current tool name.
- **Plans for Customer Communication**
    - No customer communication needed. Documentation/skill-content change only.
- **Impact Analysis**
    - Affects agents that load the `dataapp-developer` skill when creating/editing Streamlit data apps. Before this change they were steered to a non-existent tool; after it they call the correct one. No effect on single-tenant setups or runtime behavior of the MCP server itself.
    - Not behind a feature flag.
- **Deployment Plan**
    - Merges to `main` in the `ai-kit` marketplace; picked up by agents on the next skill sync. No staged rollout needed.
- **Rollback Plan**
    - Fully reversible by reverting the commit. No one-way-door change.
- **Post-Release Support Plan**
    - No special monitoring required. Support team does not need to be notified.

### DIFF
--- a/plugins/dataapp-developer/skills/dataapp-development/SKILL.md
+++ b/plugins/dataapp-developer/skills/dataapp-development/SKILL.md
@@ -35,7 +35,7 @@ If unsure → `references/choosing-app-type.md`. Short version:
 
 `references/deployment-paths.md` covers all three:
 
-- **Path A — Claude Desktop / web (MCP-only, no filesystem):** Use `modify_data_app` / `deploy_data_app` MCP tools (Streamlit only today).
+- **Path A — Claude Desktop / web (MCP-only, no filesystem):** Use `modify_streamlit_data_app` / `deploy_data_app` MCP tools (Streamlit only today).
 - **Path B — Claude Code / local agent with filesystem + MCP:** Edit files locally, push to customer git, deploy via MCP or kbagent.
 - **Path C — CLI agent (`kbagent`):** Full lifecycle via `kbagent data-app` command group.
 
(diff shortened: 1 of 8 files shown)


AdamVyborny (dismissed):

AdamVyborny:
Not sure if plugin version needs to be updated

AdamVyborny (approved):

PR #83, opened 2026-07-08
feat(dataapp-developer): add mcp-data-app skill

jordanrburger:
## What

Adds a new **`mcp-data-app`** skill to the `dataapp-developer` plugin: host any MCP server as a private, single-tenant Keboola data app. Based on [`keboola-rnd/mcp-data-app-example`](https://github.com/keboola-rnd/mcp-data-app-example), with the Keboola MCP server as the built-in worked example.

## Contents

- `SKILL.md` — router: trigger description + 6-step workflow (choose server → scaffold → secrets → git → deploy → connect)
- `reference/` — `deploy.md`, `auth-and-clients.md`, `adapting-to-any-server.md`, `troubleshooting.md`
- `scripts/scaffold.sh` — copies the bundled template into a target repo
- `template/` — the deployable app: Starlette wrapper (static-bearer + OAuth-shape client auth) + `keboola-config/` trio (nginx SSE/DNS-rebind-safe, supervisord, setup.sh), `pyproject.toml` pinned to upstream `v1.73.1`

## Key behaviors

- Deploys as a **Python/JS** app from git via **kbagent**, **keboola-git** (managed Forgejo), or the **hosted Keboola MCP tools (incl. Kai)**; manual Keboola UI fallback.
- App-level auth must be **None** (platform OIDC strips `Authorization`); `#MCP_API_KEY` is the security boundary.
- Two client-connect patterns: static bearer and claude.ai OAuth-shape connector; Kai documented as both a deploy driver and a client.

## Housekeeping

- `dataapp-developer` bumped 1.3.0 → **1.4.0** (plugin.json + marketplace.json), plugin + root README updated.
- `claude plugin validate .` passes.


plugins/dataapp-developer/skills/mcp-data-app/SKILL.md
```diff
@@ -0,0 +1,68 @@
+---
+name: mcp-data-app
+description: >-
```
davidesner, line 3: I miss some use cases. Why should the agent choose to run MCP server as Keboola data app. Not sure if it belongs here or eventually in some dedicated use case list. The intention is that the agent (Kai) can eventually suggest doing this in the right situations because users might not think of it themselves.

plugins/dataapp-developer/skills/mcp-data-app/SKILL.md
```diff
@@ -0,0 +1,68 @@
...
+  Use this skill to host an MCP server as a private, single-tenant Keboola
```
tomasfejfar: shorter

tomasfejfar: eats too much context

tomasfejfar (approved):
Except the description that is too long!

plugins/dataapp-developer/skills/mcp-data-app/SKILL.md
jordanrburger: Good call — added a **"When to use this — and when to suggest it"** section in the SKILL.md body (commit a1c35ca) with concrete use cases: no hosting for an MCP server, needing a dedicated/isolated instance vs. the shared `mcp.<region>.keboola.com` endpoint, self-hosting/forking to add custom tools, plus an explicit "proactively suggest it (e.g. as Kai) when the situation fits" framing so agents surface it when users wouldn't think of it. Kept it in the body (loads only on activation) rather than the always-loaded description.

plugins/dataapp-developer/skills/mcp-data-app/SKILL.md
jordanrburger: Trimmed — the frontmatter `description` is roughly halved (~830 → ~470 chars) in commit a1c35ca. Moved the implementation detail and use cases into the SKILL.md body (which only loads when the skill activates), so the always-loaded description no longer eats context.

PR #94, opened 2026-08-03
fix(sl-toolkit): edits use PATCH, and sql_dialect is snake_case

cjayyy:
Two bugs in the `semantic-layer` skill, both verified against a live metastore
(canary-orion, project 431). Either one alone is enough to make the skill fail or do damage.

Found while porting this skill into Kai's in-platform tooling
([AI-3661](https://linear.app/keboola/issue/AI-3661/kai-can-build-and-maintain-the-semantic-layer-bulk-createupdatedelete)) —
rather than working around them downstream, fixing them here.

## 1. `sqlDialect` is rejected — `/sl-build` cannot create a model today

The API requires snake_case `sql_dialect`. camelCase gives:

```
422  missing property 'sql_dialect'
```

The skill's POST envelope, `sl-build`'s model-create call, and the test fixture all used
`sqlDialect`, so the greenfield wizard fails at its very first write.

**Both CI guards were asserting the wrong direction.** `test_sqldialect_is_camelcase` and
`test_no_dialect_drift` *required* camelCase and *forbade* the correct key, and
`schemas/semantic-model.json` did the same in its `not.anyOf`. That's why the bug shipped and
stayed green. Inverted them, and the consistency check now matches the **quoted payload key**
instead of the bare word, so the gotchas section can still name `sqlDialect` in prose to say
it's rejected.

Also corrected the schema enum — only `Snowflake` and `BigQuery` are accepted. `Redshift` and
`Postgres` were listed; both return `422 value must be one of 'Snowflake', 'BigQuery'`.

## 2. `PATCH` exists — the documented DELETE + POST edit path is destructive

The skill stated *"The metastore has no PATCH — editing is DELETE old + POST updated"* and built
rollback machinery around it. `PATCH /api/v1/repository/{type}/{id}` works:

| | before | after |
|---|---|---|
| uuid | `019fc788-d6cb-7582-95d5-46d96f1bad9f` | **same** |
| revision | 1 | **2** |

Verified on both a `semantic-metric` and a `semantic-glossary` object, so it isn't type-specific.

Delete-then-post instead mints a **new UUID**, resets revision history, breaks anything
referencing the old UUID, and leaves the layer missing an object if the POST fails.

Rewrote the edit section around a new `api_patch` helper — send only the fields that change. The
rename cascade to constraint `metrics[]` is preserved and is now also in-place. The rollback block
is deleted: nothing is removed, so there is nothing to roll back.

## Testing

All 30 tests pass. I also re-introduced the `sqlDialect` bug in the fixture and confirmed the
corrected guards now catch it (2 failures) rather than demanding it.

```
30 passed in 0.06s
```

## Release Notes

Fixes the `semantic-layer` skill's model-create payload (`sql_dialect`, not `sqlDialect`), which
was rejected by the metastore with a 422, and replaces the destructive DELETE + POST edit
recipe with in-place `PATCH` so object UUIDs and revision history survive an edit.

### DIFF
--- a/plugins/sl-toolkit/README.md
+++ b/plugins/sl-toolkit/README.md
@@ -42,7 +42,7 @@ Supports GCP, AWS, and Azure stacks — metastore region derived automatically f
 
 ## Key design decisions
 
-**No PATCH endpoint** — all edits are DELETE old id + POST new, with rollback on POST failure.
+**Edits use PATCH** — `PATCH /api/v1/repository/{type}/{id}` updates in place, preserving the object's UUID and bumping its revision. Never DELETE + POST: that mints a new UUID, resets revision history, and breaks anything referencing the old one.
 
 **Constraint cascade on rename** — renaming a metric auto-updates constraint `metrics[]` references to prevent orphan FKs in `DIM_METRIC_THRESHOLD` tables.
 
--- a/plugins/sl-toolkit/commands/sl-build.md
+++ b/plugins/sl-toolkit/commands/sl-build.md
@@ -320,7 +320,7 @@ if UPDATE_ID:
 else:
     uuid = api_post('/api/v1/repository/semantic-model', {
         'name': model['name'],
-        'data': {'name': model['name'], 'description': model['description'], 'sqlDialect': 'Snowflake'},
+        'data': {'name': model['name'], 'description': model['description'], 'sql_dialect': 'Snowflake'},
         'branch': 'main', 'schemaVersion': '1.0.0', 'scope': 'project'
     })['data']['id']
     print(f'✓ model created {uuid}')
--- a/plugins/sl-toolkit/skills/semantic-layer/SKILL.md
+++ b/plugins/sl-toolkit/skills/semantic-layer/SKILL.md
@@ -84,6 +84,12 @@ def api_post(path, body):
     with urllib.request.urlopen(req, timeout=30) as r:
         return json.loads(r.read())
 
+def api_patch(path, body):
+    req = urllib.request.Request(
+        f"{METASTORE}{path}", json.dumps(body).encode(), H, method='PATCH')
+    with urllib.request.urlopen(req, timeout=30) as r:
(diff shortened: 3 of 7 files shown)


cjayyy:
CI note: **Tier 0 (static lint + case validation) passes** — that's the job covering these changes, and it's the one that was previously asserting the wrong casing.

**Tier 1 (skill activation routing) fails for an unrelated infra reason** — the repo's Anthropic key is out of credit:

```
anthropic.BadRequestError: Error code: 400 — 'Your credit balance is too low to
access the Anthropic API. Please go to Plans & Billing to upgrade or purchase credits.'
```

It errors before evaluating any case, so it's not a signal about this diff. Locally: `30 passed`.

cjayyy:
Re-checked the Tier 1 failure after the CI notification — **it is not caused by this PR, and nothing in the diff can fix it.** Evidence:

1. **It fails on the API call, before evaluating any case.** The job runs 3s and dies at:
   ```
   anthropic.BadRequestError: Error code: 400 — 'Your credit balance is too low to
   access the Anthropic API. Please go to Plans & Billing to upgrade or purchase credits.'
   ```
   The secret *is* present (the `secret not available` line in the log is the script being echoed, not the branch taken), so this isn't a fork-skip — the repo's Anthropic account is out of credit.

2. **Tier 1 routes on skill frontmatter, which this PR does not touch.** `git diff origin/main..HEAD -- …/SKILL.md` contains no `name:` or `description:` change.

3. **The deterministic half of Tier 1 passes locally** — `activation/test_activation_cases.py` + `test_parse_answer.py`: **80 passed**. Those are the offline case-set and answer-parsing checks that don't need an API key.

4. **Tier 0 — the job that actually covers these changes — passes**, and it's the one that was previously asserting the wrong `sql_dialect` casing.

So: needs credit topped up on the repo's Anthropic account to go green. Happy to re-run once that's sorted.

Matovidlo (approved):

PR #98, opened 2026-08-18
docs(dataapp-developer): document Storage Access enablement and fix stale claims

MiroCillik:
Fixes documentation in the `dataapp-development` skill that was wrong or missing, found by building and deploying a real Python/JS data app end-to-end. Everything here was verified against a live app (project 40, `connection.eu-central-1.keboola.com`) or read from the source of the tool being described — no inferred behavior.

Refs [AJDA-3183](https://linear.app/keboola/issue/AJDA-3183/), [AI-3753](https://linear.app/keboola/issue/AI-3753/), [AI-3751](https://linear.app/keboola/issue/AI-3751/), [CFTL-716](https://linear.app/keboola/issue/CFTL-716/).

> **Scope reduced.** This PR originally added a drafts/previews recipe. Removed in `ee78053` per @davidesner's review — the drafts & deployments API is under active development ([AJDA-3034](https://linear.app/keboola/issue/AJDA-3034/), [AJDA-3140](https://linear.app/keboola/issue/AJDA-3140/)), and documenting today's manual mechanism would entrench a pattern that keeps `main` writeable and decouples app deployment from branch merge. What remains is unrelated to drafts.

## `storage-access.md` — new §Enabling Storage Access

`runtime.workspace.enabled: true` is what makes the platform provision the app's ephemeral workspace and inject `WORKSPACE_ID` / `QUERY_SERVICE_URL` / `KBC_WORKSPACE_MANIFEST_PATH`. It appeared **nowhere** in the skill, yet every storage pattern the skill documents depends on it — including DuckDB-cached RO, which `duckdb-caching.md` presents as *the* default for read-only dashboards.

The failure mode is the expensive kind: the app builds, deploys, reports `state=running`, passes its health probe — and reads zero rows. Confirmed in both directions on a live app: absent → `Missing env vars: WORKSPACE_ID` while every platform signal stayed green; present → full cache load.

**The prerequisite has since landed**, which reshaped this section. [keboola/cli#626](https://github.com/keboola/cli/pull/626) gave `kbagent data-app create` a `--workspace / --no-workspace` flag **defaulting ON**, released in kbagent 0.87.0. So the section no longer teaches a hand-edit as the normal path — it states that every creation path sets the flag for you, and that the only decision left is the reverse one: pass `--no-workspace` for an app that never touches Storage. That is what @davidesner's review asked for. The silent-failure warning stays, because it is still reachable — apps created on ≤ 0.86.0 predate the default, and `--no-workspace` opts out of it.

## `duckdb-caching.md` — prerequisite line

One line naming the flag in the pattern's prerequisites. This is [AJDA-3183](https://linear.app/keboola/issue/AJDA-3183/) item 3, which cli#626 explicitly left to this repo. It matters here specifically because the skill makes DuckDB-cached RO *the* default for read-only dashboards, so this is where an agent is most likely to meet the silent empty dashboard.

## `troubleshooting.md`

* The existing Storage Access env-var entry gave a **UI-only** fix and conflated enabling access with adding *writable* tables (`unload_strategy: "direct-grant"`), which a read-only app doesn't need. Rewritten for the read-only case, it now names why a modern app could still lack the flag (created pre-0.87.0, created with `--no-workspace`, or UI Advanced Settings left off), and it states that the app still reports healthy while broken.
* New entry: **config change has no effect after redeploy**, with the `config_version_storage` vs `config_version_deployed` diagnostic. Hit this live — a config change silently didn't deploy across a warm redeploy *and* a cold stop→deploy; only an explicit `--config-version` applied it.
* §Reading logs: `kbagent data-app logs` **has shipped** (verified on 0.84.2) — the skill still described it as an unshipped follow-up. Now documented, along with the rough edges found in use: `400 ... is not running` on a stopped app, the manage-token requirement on some subcommands, and the 20-line MCP tail cap.

### A claim this PR introduced, then corrected

An earlier revision said `--config-version` *"claims to default to latest; it does not."* **That was wrong** and is fixed in the final diff. Since 0.65.0 `data-app deploy` does resolve the latest Storage version and pin it. Reading `services/data_app_service.py` on cli `main`, the real behavior is narrower:

* **Streamlit / external git** (`parameters.dataApp.git` present in the config) → deploy resolves the latest version and pins it. Config changes deploy normally.
* **Pure managed git** (`hasManagedGitRepo` and no `dataApp.git` block — exactly what `--use-managed-git-repo` produces) → deploy **deliberately omits** `configVersion`, because the source resolves through `app.managedGitRepoId` and pinning a git-blockless config makes the runtime demand `dataApp.git.repository`. The side effect is that nothing advances the operator's pin, so `config.json` changes never reach the container.

So it was a real bug in a specific app shape, not a lying default. [CFTL-716](https://linear.app/keboola/issue/CFTL-716/) (still In Progress) is the platform-side half: the operator keys the pod restart on the pin, not on the config's content. The entry now explains both branches and keeps `--config-version` as the documented escape hatch.

## `python-js-apps.md` — §Deployment from Keboola-managed git

Replaces the `Deployment via MCP — PLACEHOLDER` section, which claimed customer-provided git was *"the only supported path"*. Managed Forgejo repos work today via `--use-managed-git-repo`, and MCP `modify_python_js_data_app` covers Python/JS create/update; documented with pointers to the `keboola-git` skill for the repo plumbing.

## `deployment-paths.md`

Three more instances of the same wrong "no Python/JS via MCP" claim, corrected to name the right tool per app type. One of them also dangled at the section this PR renames.

## `TODO.md`

* [AJDA-2924](https://linear.app/keboola/issue/AJDA-2924/) (`workspace.enabled=true` by default at the platform level) was **canceled 2026-08-11**, so the config flag stays explicit — but both creation paths now write it, so the entry is reframed as resolved at the tooling layer rather than pending.
* Dropped a stale claim that a *"flip workspace on first"* caveat lives in `SKILL.md` and `deployment-paths.md`. It is in neither (`grep` → 0 hits); it was planned and never landed.
* `data-app logs` marked shipped; the `data-app create` workspace gap marked shipped; the `--config-version` gap replaced with the accurate managed-git one.

## Version bump

`1.5.1` → `1.5.2` in `plugins/dataapp-developer/.claude-plugin/plugin.json` and `.claude-plugin/marketplace.json`, per `CLAUDE.md`.

## Release Notes

* **Justification**
  * Agents building Keboola data apps were being misled by the skill's own docs. The one config flag that makes Storage reads work was never named, so an app would deploy green and silently serve no data — the most expensive kind of wrong, because every signal reports success. A stale claim also told agents a shipped command didn't exist.
  * For a non-technical consumer: the guide Claude follows when building a Keboola data app now explains how an app gets access to project data, and how to tell when a change didn't actually deploy.
* **Plans for Customer Communication**
  * None needed — documentation only, no product surface change. Worth mentioning to anyone who has hit "my data app runs but shows no data", which now has a documented cause and fix.
* **Impact Analysis**
  * Docs-only change to one plugin's skill content. No code, no runtime, no single-tenant impact. Not behind a feature flag; applies to everyone who installs the plugin.
  * Risk is limited to the guidance itself being wrong. Each claim was verified against a live app or read from the source of the tool described. The two platform issues referenced ([CFTL-716](https://linear.app/keboola/issue/CFTL-716/), [AJDA-2924](https://linear.app/keboola/issue/AJDA-2924/)) are linked so the text can be revisited when they change.
* **Deployment Plan**
  * Continuous — merge to `main` and it ships with the plugin version bump in this PR. Not critical.
* **Rollback Plan**
  * Fully reversible: revert the commits. No state, no migration, two-way door.
* **Post-Release Support Plan**
  * No support burden. Two sections will need revisiting when upstream work lands: the redeploy entry once [CFTL-716](https://linear.app/keboola/issue/CFTL-716/) is fixed, and the managed-git section once the drafts & deployments API ships.
  * Support team does not need notifying.

## Verification

* `runtime.workspace.enabled` cause and effect confirmed both directions on a live app; data path proven from container logs (`cached 2412 beers`), not from `state=running`.
* `--workspace / --no-workspace` confirmed present and defaulting to `True` in cli `main` (`commands/data_app.py`), released in 0.87.0 per its own changelog.
* Deploy's config-version resolution branches read from `services/data_app_service.py`; CFTL-716 confirmed still In Progress, so that entry stands.
* [CFTL-716](https://linear.app/keboola/issue/CFTL-716/) reproduced: warm redeploy and cold stop→deploy both left the deployment pinned to the stale version; explicit `--config-version` applied it. Reported on that issue.
* `kbagent data-app logs` exercised throughout.
* Also verified `config update --merge` **preserves** `parameters.dataApp.git` — which cleared `config update` of suspicion for an earlier breakage; the real cause is `create --use-managed-git-repo` not persisting `dataApp.git`, recorded in `TODO.md`.
* All internal markdown links in changed files resolve; zero `linear.app` links in `references/`; Tier 0 evals green locally (330 passed, 32 skipped) and on CI.

## Review

* Copilot raised 5 comments; 4 applied, 1 rejected with evidence (it flagged `kbagent data-app logs` as unavailable, citing the skill's own stale text — the command exists, so the stale text was fixed instead). Its version-bump concern is addressed in this round.
* @davidesner requested changes; the drafts scope was dropped, the Linear links removed from `references/`, the MCP Python/JS claim corrected, and the Storage Access section rewritten around the CLI default now that AI-3753 has landed. Replies are in the threads.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)


MiroCillik:
Note: This is for the current drafts implementation - NOT for the NEW ONE - which is under development. 
But this is a pain for some important clients who rely heavily on local/MCP apps development.

plugins/dataapp-developer/skills/dataapp-development/references/drafts-and-previews.md
```diff
@@ -0,0 +1,303 @@
...
+```
+
+Anything printed must stop being tracked and move into `keboola-config/setup.sh` (build-at-deploy). Then push:
+
+```bash
+SECRET=$(mint_secret)
+[ -n "$SECRET" ] || { echo "credential mint failed"; exit 1; }
+git push "https://kai:${SECRET}@git.${STACK}/keboola/app-${PROD_APP_ID}.git" "$BRANCH:$BRANCH" 2>&1 \
+  | sed "s/${SECRET}/<redacted>/g"
+unset SECRET
```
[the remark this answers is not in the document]
MiroCillik: Fixed, at a different layer than suggested. Literal-string replacement still leaves the secret in the URL, and `git` echoes the remote URL in its own error output — so the leak channel stays open and we just get better at scrubbing it. Removed the secret from the URL entirely instead, via a credential helper:

```bash
git_push() {   # usage: git_push <refspec>
  git -c credential.helper='!f() { test "$1" = get && printf "username=kai\npassword=%s\n" "$GITPAT"; }; f' \
      push "https://git.${STACK}/keboola/app-${PROD_APP_ID}.git" "$1"
}
```

Verified against the live managed repo (`ls-remote` and `push` both authenticate). Nothing sensitive in the URL now, so a failed push has nothing to leak and no redaction step is needed.

On the regex point specifically: I minted a credential and classified it without printing — 40 chars, `[a-z0-9]` only, no metacharacters and no `/`. So the concern would not have fired on the current token format. It is still the right call, since that charset is not a documented guarantee, and the failure mode is a leaked credential. Both push sites updated, and the doc now explains why URL-embedding is avoided rather than just saying "redact".

plugins/dataapp-developer/skills/dataapp-development/references/drafts-and-previews.md
```diff
@@ -0,0 +1,303 @@
...
+```
+
+**`--config-version` is not optional in practice.** Its help text says *"defaults to latest"*; it does not. Omitting it leaves the deployment pinned to the previous version — the step-3 changes silently never take effect and the deploy still reports `state=running`. A cold `stop` → `deploy` does **not** work around it. This is [CFTL-716](https://linear.app/keboola/issue/CFTL-716/) (root-caused in keboola-operator).
+
+## Step 5 — verify the draft, and verify production is untouched
+
+`state=running` is not proof of anything. Neither is an HTTP probe: an unauthenticated `GET /` returns the **platform login gate** (HTTP 200) rather than the app, and `POST /` returns 200 from the nginx health rule regardless of app state. **The logs are the authoritative signal.**
+
+```bash
+kbagent data-app logs --project "$PROJECT" --app-id "$DRAFT_APP_ID" --lines 200
```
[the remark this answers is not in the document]
MiroCillik: Not fixing this one — the premise is out of date, and the stale text is in the repo rather than in this PR.

`kbagent data-app logs` exists and works:

```
$ kbagent data-app logs --help
Usage: kbagent data-app logs [OPTIONS]
  Tail the container logs for a deployed data app.
```

I used it throughout building and deploying the app this reference documents; it is how I confirmed the draft was reading data (`[duck] cached 2412 beers`). It is also the *most complete* log view outside the UI — the MCP tail is capped at 20 lines.

So rather than weakening the reference with a fallback, I fixed what was actually wrong: `troubleshooting.md` §Reading logs and `TODO.md` both still described the command as an unshipped follow-up. Both now document it, plus the rough edges I hit in real use — `400 ... is not running` when the app is stopped, the manage-token requirement on some subcommands, and the JSON-per-line hard wrapping that makes grepping awkward.

Worth flagging that this is the same failure mode the PR exists to fix: the reviewer reasoned from the repo docs, and the repo docs were stale.

plugins/dataapp-developer/skills/dataapp-development/references/storage-access.md
```diff
@@ -88,6 +89,41 @@ For every other case → omit `BRANCH_ID` (or set it to `default`) and the app r
...
+`KBC_WORKSPACE_MANIFEST_PATH` at deploy. A **read-only** app needs nothing else — no input
+mapping, no output mapping.
+
+Set it via kbagent (UI: App configuration → Advanced Settings → Storage Access):
+
+```bash
+kbagent config update --project P --component-id keboola.data-apps --config-id CFG --merge \
+  --set 'runtime.workspace.enabled=true'
+# then redeploy PINNED to the new version — a plain redeploy will not pick it up:
+kbagent data-app deploy --project P --app-id N --config-version <latest> --wait
```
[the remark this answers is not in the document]
MiroCillik: Fixed. Replaced with the capture idiom that `drafts-and-previews.md` already uses, so the snippet is executable as-is:

```bash
VERSION=$(kbagent --json data-app detail --project P --app-id N \
  | python3 -c "import sys,json;print(json.load(sys.stdin)[\"data\"][\"config_version_storage\"])")
kbagent data-app deploy --project P --app-id N --config-version "$VERSION" --wait
```

plugins/dataapp-developer/skills/dataapp-development/references/troubleshooting.md
```diff
@@ -78,14 +79,41 @@
...
+
+Note the app **still deploys, reports `state=running`, and passes its health probe** — it just cannot read data. This line in the container log is the only signal.
 
 **Fix:**
-- **Production:** UI → Advanced Settings → enable Storage Access and add writable tables with `unload_strategy: "direct-grant"`. Redeploy.
+- **Production:** set `runtime.workspace.enabled=true` on the config, then redeploy **pinned to the new version**:
+  ```bash
+  kbagent config update --project P --component-id keboola.data-apps --config-id CFG --merge \
+    --set 'runtime.workspace.enabled=true'
+  kbagent data-app deploy --project P --app-id N --config-version <latest> --wait
```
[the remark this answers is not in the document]
MiroCillik: Fixed the same way as `storage-access.md`. This file had both variants — `<latest>` on one line and `<config_version_storage>` on another — so I standardised both on the executable capture form. Reading the version fresh also matters beyond copy/paste safety: reusing a version captured earlier can deploy a stale config if anything touched the configuration in between, which is precisely the CFTL-716 failure this section warns about.

plugins/dataapp-developer/skills/dataapp-development/TODO.md
```diff
@@ -20,7 +21,7 @@ Gaps that force the skill to recommend kbagent or filesystem paths for things MC
 - **No log-reading tool beyond the 20-line tail.** `get_data_apps(...).deployment_info.logs` returns only the most recent lines. For real debugging the agent has to direct the user to the Keboola UI Terminal Log tab.
 - **No workspace management.** Cannot create / grant / list / delete workspaces via MCP. Local-dev workspace setup falls back to UI or kbagent.
 - **No direct secret management.** Adding / removing / listing `dataApp.secrets` has to go through `modify_streamlit_data_app` (which couples it to source-code edits) or the Configuration API directly.
-- **No Storage Access toggle.** Enabling Storage Access on an app config requires UI navigation; no MCP affordance.
+- **No Storage Access toggle.** No MCP affordance for `runtime.workspace.enabled`. Workaround is `kbagent config update … --set 'runtime.workspace.enabled=true'` (documented in `storage-access.md` §Enabling Storage Access), or UI navigation.
```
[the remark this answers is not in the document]
MiroCillik: Fixed — dangling reference introduced by this PR. Now points at `references/python-js-apps.md` §Deployment from Keboola-managed git, which is the section that replaced the placeholder. The MCP-specific gap it was illustrating (`modify_streamlit_data_app` being Streamlit-only) is unchanged and still listed.

davidesner:
@MiroCillik I thought we would wait for the new API. Isn't it available already? So things like

> Kai's "Publish to Production" is: merge the branch into `main`, redeploy production, then remove the draft.` 

are not necessary. Leaving the main writeable and the deploy mechanism decoupled between app deployment and branch merge doesn't seem like a good pattern to me.

MiroCillik:
Yeah, I guess it's better to wait for the new API regarding drafts.. But there are also other improvements/fixes in this PR, I will remove the "drafts" part from it.

MiroCillik:
Agreed with @davidesner — dropped the drafts/previews part in `ee78053`. The objection is the right one: a merge-to-`main`-then-redeploy publish step keeps `main` writeable and decouples app deployment from branch merge, and writing that into the skill would have entrenched exactly the pattern the new drafts & deployments API ([AJDA-3034](https://linear.app/keboola/issue/AJDA-3034/), [AJDA-3140](https://linear.app/keboola/issue/AJDA-3140/)) is replacing. Better an empty section than a blessed workaround.

**Removed**

- `references/drafts-and-previews.md` (whole file)
- both `SKILL.md` router entries and the two preview/draft trigger evals
- the preview recipe in `python-js-apps.md` — now a short note stating it is intentionally absent, why, and pointing at the UI flow (Kai's draft + "Publish to Production") meanwhile
- the `TODO.md` draft-command gaps, reframed as *"waiting on the new API, not on kbagent"* so nobody picks up a `data-app draft` CLI task that the new API supersedes

**Kept** — none of this is draft-related:

| Change | Why |
|---|---|
| `storage-access.md` §Enabling Storage Access | `runtime.workspace.enabled` appeared nowhere in the skill, yet every workspace query needs it. Without it an app deploys, reports `state=running`, passes its health probe, and reads zero rows ([AI-3753](https://linear.app/keboola/issue/AI-3753/)) |
| `troubleshooting.md` | read-only Storage Access fix (the old entry gave a UI-only, read-write `direct-grant` answer); new stale-config-after-redeploy entry ([CFTL-716](https://linear.app/keboola/issue/CFTL-716/)); corrected §Reading logs |
| `python-js-apps.md` §Deployment from Keboola-managed git | replaces a PLACEHOLDER that wrongly said customer-provided git was "the only supported path" ([AI-3751](https://linear.app/keboola/issue/AI-3751/)) |
| `TODO.md` | AI-3218 was canceled as [AJDA-2924](https://linear.app/keboola/issue/AJDA-2924/), so the workspace flag is permanent rather than pending; dropped a stale claim about a caveat that exists in neither `SKILL.md` nor `deployment-paths.md`; marked `data-app logs` shipped |

Diff is now 4 files, +104/−16, no new files. Retitled the PR to match.

The Copilot fixes from the previous round survive where they applied to kept files — the `<latest>` → captured-`config_version_storage` change in `storage-access.md` and `troubleshooting.md`, and the dangling `TODO.md` cross-reference. The credential-redaction fix went away with the file it was in.

One consequence worth naming: [AI-3754](https://linear.app/keboola/issue/AI-3754/) asked for a `kbagent data-app draft` command group. If the new API is the path, that ask is superseded and the issue should probably be reduced to the two findings that outlive it — `--use-managed-git-repo` can't target an existing app's repo, and `data-app create --use-managed-git-repo` doesn't persist `parameters.dataApp.git`. Happy to rewrite it that way; also happy to leave it for whoever owns the new API to fold in.

plugins/dataapp-developer/skills/dataapp-development/references/python-js-apps.md
```diff
@@ -291,12 +291,26 @@ Exit code **153** means the locked commit no longer exists in the remote (force-
...
+
+**`--use-managed-git-repo` only provisions a NEW empty repo.** It cannot point an app at an existing app's repo. A second app sharing one repo has to be configured as "external git" against the managed URL, with its own minted credential.
 
-**Status today: not yet finished.** Agents working on Python/JS apps fall back to **customer-provided git** (private GitHub/GitLab with PAT or SSH key) as the only supported path.
+**Branch previews / drafts: deliberately not documented here.** Previewing a branch on its own URL before publishing is possible today by hand, but the mechanism — a second app configuration on the same repo, plus a merge-to-`main`-then-redeploy publish step — is being replaced by a dedicated drafts & deployments API ([AJDA-3034](https://linear.app/keboola/issue/AJDA-3034/), [AJDA-3140](https://linear.app/keboola/issue/AJDA-3140/)). Writing the current workaround into the skill would entrench a pattern that keeps `main` writeable and decouples app deployment from branch merge, so this section stays empty until the new API lands. Until then, use the Keboola UI (Kai's draft + "Publish to Production" flow) rather than assembling it from CLI primitives.
```
davidesner: Not sure this is necessary in the skill itself. maybe leave unchanged for now?

plugins/dataapp-developer/skills/dataapp-development/references/python-js-apps.md
```diff
@@ -291,12 +291,26 @@ Exit code **153** means the locked commit no longer exists in the remote (force-
...
 
-When the platform support lands, this section expands. If it grows past ~50 lines, split it into its own reference.
+**Not yet available via MCP:** `modify_streamlit_data_app` is Streamlit-only, so Python/JS app creation and deployment go through `kbagent` (or the Configuration API) rather than MCP tool calls. Check the tool surface before assuming either way — availability varies by stack and MCP server version:
```
davidesner: Technically the MCP does have the modify_python_js_data_app tool.

plugins/dataapp-developer/skills/dataapp-development/references/storage-access.md
```diff
@@ -88,6 +89,43 @@ For every other case → omit `BRANCH_ID` (or set it to `default`) and the app r
 - `branch_id` — the numeric ID to paste into `.env.local`.
 - `is_development_branch` — confirms which branch the MCP session is currently scoped to. **Must be `false`** before relying on `branch_id`. If `true`, the MCP is in a dev-branch context — switch to the production branch in your MCP setup and re-run, otherwise you'll paste a dev-branch ID into `.env.local` and the app will read dev-branch tables locally.
 
+## Enabling Storage Access (required for ANY workspace query)
```
davidesner, line 92: @MiroCillik 

I think this `**`kbagent data-app create` does not set this flag**, and neither does the UI's app-creation
flow in every case. Check it before the first deploy.` should be changed and this option exposed via flag explicitly, rather than adding it directly like this (more prone to error) + this functionality should work with CLI alone. The same goes for https://github.com/keboola/mcp-server/blob/main/src/keboola_mcp_server/tools/data_apps.py#L1298 MCP tool. I leave up to you whether this should be a followup action or whether it should supersede this.

To be fair since every pattern listed requires this (i.e. every data app that requires access to storage) the ideal state would be that CLI and MCP tools created it by default with this flag by default. (I know you  said it is a security risk https://linear.app/keboola/issue/AJDA-2924/dataapps-workspaceenabledtrue-by-default-for-every-new-app). Then this entire section could be reduced to a statement that when the app doesn't require storage access, the workspace access should be explicitly disabled.

plugins/dataapp-developer/skills/dataapp-development/references/storage-access.md
```diff
@@ -88,6 +89,43 @@ For every other case → omit `BRANCH_ID` (or set it to `default`) and the app r
...
+```
+
+**`kbagent data-app create` does not set this flag**, and neither does the UI's app-creation
+flow in every case. Check it before the first deploy.
+
+**The failure mode is silent.** Without the flag the app still builds, still reports
+`state=running`, and still passes its health probe — it simply cannot read any data. The only
+signal is the app's own startup error in the container log
+(`Missing env vars: WORKSPACE_ID`). Verify from the logs, not from `state`. Tracked in
+[AI-3753](https://linear.app/keboola/issue/AI-3753/).
```
davidesner: I don't think having linear references in public skill is helpful (also [here](https://github.com/keboola/ai-kit/blob/ee78053e1bad33907250543d71b8bfd1727fe236/plugins/dataapp-developer/skills/dataapp-development/references/python-js-apps.md?plain=1#L310))

davidesner (changes requested):

plugins/dataapp-developer/skills/dataapp-development/references/storage-access.md
MiroCillik: Yeah, the linear references are very weird

plugins/dataapp-developer/skills/dataapp-development/references/storage-access.md
MiroCillik: I agree, this should be a default in CLI and MCP and not need a special config update call -> will create follow up issues

plugins/dataapp-developer/skills/dataapp-development/references/python-js-apps.md
MiroCillik: Dropped. The paragraph is gone, and so is every other reference to drafts/previews in the skill — the whole section now covers only managed-git provisioning.

For the record on why I had added it: my worry was that an agent finding nothing would reinvent the manual workaround from primitives. Your call stands though, and there is a better answer to that worry now — MCP `modify_python_js_data_app` handles drafts directly (`is_draft`, `parent_configuration_id`, `branch`), so an agent has a real path rather than a gap to fill in. Nothing for the skill to say here.

plugins/dataapp-developer/skills/dataapp-development/references/python-js-apps.md
MiroCillik: Correct, and my claim was wrong. Verified in `mcp-server/src/keboola_mcp_server/tools/data_apps.py:879` — `modify_python_js_data_app` exists and covers create/update, managed-git provisioning (returns `repo_url`) and drafts.

Where my wrong impression came from, in case it is worth fixing on the kbagent side: `kbagent tool call modify_python_js_data_app` returns `Unknown MCP tool` on 0.84.2, because the bundled `keboola-mcp-server` is older than the tool. That is a kbagent packaging lag, not an MCP gap, and I misread it as the latter.

Fixed in `python-js-apps.md`, and I found the same false claim in **three more places** in `deployment-paths.md` that your comment led me to:

- §Path A limitations — *"No Python/JS type via MCP today (planned — see … PLACEHOLDER)"*, which also dangled at the section this PR renames
- §Path B — *"Today this means using `kbagent` (Path C) — MCP doesn't yet support Python/JS app deployment"*
- §Decision shortcuts — *"Path C (until MCP gains support)"*

All three corrected to name the right tool per app type.

plugins/dataapp-developer/skills/dataapp-development/references/storage-access.md
MiroCillik: Reworked, and treating the tooling change as follow-up rather than superseding this — but your comment turned up more than the framing being wrong, so worth spelling out what I found in `data_apps.py`:

- **MCP already sets the flag.** Line 1242: `runtime=(CodeDataAppConfig.Runtime(workspace=…Workspace(enabled=True)) if has_storage_workspace else None)`.
- **It is gated on a project feature** I did not know existed — line 154, `DATA_APPS_STORAGE_WORKSPACE_FEATURE = 'data-apps-storage-workspace'`.
- **There is a fallback when the feature is off** — line 1113 injects `WORKSPACE_ID` through `parameters.dataApp.secrets` instead.

So my text was wrong about MCP and silent on both the feature gate and the fallback. It now states plainly who sets the flag (MCP yes, `kbagent data-app create` no, UI via Advanced Settings) and mentions the feature gate, instead of teaching a hand-edit as the normal path. Kept the silent-failure warning, because that is the part that actually costs people time — the app builds, reports `state=running`, passes its health probe, and reads zero rows.

On follow-up vs supersede: **follow-up.** The gap is real for anyone on the CLI today, so leaving it undocumented while the flag ships did not seem right. [AI-3753](https://linear.app/keboola/issue/AI-3753/) already asks for exactly the `--workspace/--no-workspace` affordance you describe; I have added the MCP-parity note there. If you would rather this section wait for the flag, I will pull it.

On defaulting it: agreed that would collapse this section to "disable it when the app does not need storage" — that is the better end state, and AJDA-2924 is where that argument belongs rather than here.

plugins/dataapp-developer/skills/dataapp-development/references/storage-access.md
MiroCillik: Removed — zero `linear.app` links in `references/` now. Fixed in both places you pointed at (`storage-access.md` and the `python-js-apps.md` paragraph, which is gone entirely), plus the `CFTL-716` link I had put in `troubleshooting.md`. Where an issue is worth naming, the text now describes the behavior instead of linking a ticket a reader outside Keboola cannot open.

`TODO.md` still has Linear links — that file already carried them on `main` (3 before this PR), and it reads as internal planning rather than agent-facing guidance. Say the word if you want those stripped too.

plugins/dataapp-developer/skills/dataapp-development/references/storage-access.md
MiroCillik: Let's wait until the AI-3753 lands and we can simplify this section

MiroCillik: @davidesner AI-3753 has landed — keboola/cli#626, shipped in kbagent **0.87.0**. `data-app create` now takes `--workspace / --no-workspace` and grants Storage access **by default**, so the hand-edit framing you objected to is gone. Your suggestion is now literally what the section does:

> Then this entire section could be reduced to a statement that when the app doesn't require storage access, the workspace access should be explicitly disabled.

The table row flipped from *"**Does not set it.** Add it before the first deploy"* to *"on by default since 0.87.0; `--no-workspace` is the opt-out"*, and the paragraph after it says the only decision left is the reverse one — pass `--no-workspace` for an app that never touches Storage. The retrofit snippet is dropped from this file entirely (it duplicated the one in `troubleshooting.md`, which is now the single place it lives). Net −6 lines here, and it reads as "this is handled for you" rather than "here is a config edit to remember".

Kept the silent-failure warning, because it is still the reachable failure: apps created on ≤ 0.86.0 predate the default, and `--no-workspace` opts out of it.

**Also in this push, one thing you did not ask for but should see** — I corrected a claim *this PR* introduced. It said `--config-version` *"claims to default to latest; it does not."* That is wrong: since 0.65.0 `deploy` resolves the latest Storage version and pins it. What I actually hit is narrower, and reading `services/data_app_service.py` on cli `main` explains it:

- **Streamlit / external git** (`parameters.dataApp.git` present) → deploy pins the latest version. Config changes deploy normally.
- **Pure managed git** (`hasManagedGitRepo`, no `dataApp.git` block — exactly what `--use-managed-git-repo` produces) → deploy **deliberately omits** `configVersion`, because pinning a git-blockless config makes the runtime demand `dataApp.git.repository`. Side effect: nothing advances the operator's pin, so `config.json` changes never reach the container.

So it was a real bug, in a specific app shape, not a lying default. CFTL-716 (still In Progress) is the platform-side half — the pod restart is keyed on the pin, not on the config content. The troubleshooting entry now explains both branches, and `TODO.md` carries the accurate version of the gap.

Two smaller things: `duckdb-caching.md` gained the prerequisite line (AJDA-3183 item 3, which cli#626 explicitly left to this repo — it is the pattern the skill makes *the* default for read-only dashboards, so it is where an agent is most likely to meet the silent empty dashboard), and the plugin version is bumped 1.5.1 → 1.5.2 in `plugin.json` + `marketplace.json` per `CLAUDE.md`, which was Copilot's one still-open concern.

Rebased onto `main`. Tier 0 green.

PR #99, opened 2026-08-19
fix(dataapp-developer): pin templates to patched query-service SDKs

Matovidlo:
## Summary
The `@keboola/query-service` JS SDK is being retired in favor of `@keboola/api-client`'s `queryService` client + `queryService` SDK (`createQueryServiceSdk`, keboola/ui#8059), which now carries the same `executeQuery`/auto-pagination fix. This repoints every JS/TS reference in the `dataapp-developer` skill accordingly:
- `templates/nodejs-app/package.json` + `api/keboola-client.js`: constructs `createQueryServiceClient(...)` then `createQueryServiceSdk({ queryServiceClient })`, calls `executeQuery` on the SDK. Also asserts `data.length === numberOfRows` after `executeQuery` as a loud-failure guard (per @MiroCillik's suggestion) — belt-and-braces in case a future SDK regression silently truncates again.
- `references/storage-access.md`, `troubleshooting.md`, `TODO.md`, `glossary.md`: every JS/TS example and mention updated to the client+SDK construction.

The Python side (`keboola-query-service`) is untouched — its own fix (keboola/query-service-api-python-sdk#9, released as v0.3.0) already ships independently and is installable today.

Also (from the original commit): bumped `templates/streamlit/pyproject.toml`'s `keboola-query-service` floor to `>=0.3.0`, and added a cache-completeness check to `duckdb-caching.md`.

Linear: [AI-3752](https://linear.app/keboola/issue/AI-3752)

## Test plan
- [x] `@keboola/api-client`'s `queryService` SDK (keboola/ui#8059) is tested there — type-check, lint, 906 tests including 10 for the SDK, all green.
- [x] Grepped this repo for any remaining `@keboola/query-service` / `query-service-api-js-sdk` / stale `client.executeQuery(...)` references — none left
- [x] `@keboola/api-client@27.0.0` is published (`npm view @keboola/api-client dist-tags` → `latest: 27.0.0`). Installed it directly and confirmed `/queryService` exports `createQueryServiceClient` and `/sdk/queryService` exports `createQueryServiceSdk`, matching the template's imports. Read the built SDK source: `executeQuery` creates the job, waits for completion, then `fetchAllResultPages` loops on `offset`/`pageSize` against `getQueryResults` until `data.length >= numberOfRows` — the auto-pagination claim in the docs holds, and `numberOfRows` is the true total (not the page count), so the template's truncation guard is non-vacuous.
- [x] Still open: an empirical run against a real >500-row table (e.g. @MiroCillik's `in.c-dbt-beers.beers`, 2412 rows) hasn't been done against v27 specifically — no KBC credentials in this environment. Static verification above is strong but not a substitute for that live check.

### DIFF
--- a/plugins/dataapp-developer/skills/dataapp-development/references/glossary.md
+++ b/plugins/dataapp-developer/skills/dataapp-development/references/glossary.md
@@ -76,7 +76,7 @@ See `references/kai-integration.md` for the embed patterns.
 
 Python SDK: https://github.com/keboola/query-service-api-python-sdk
 
-JS/TS: `@keboola/api-client`'s `queryService` client (https://github.com/keboola/ui, `packages/api-client`) — the formerly-standalone `query-service-api-js-sdk` repo is retired; its `executeQuery`/pagination behavior now lives here.
+JS/TS: `@keboola/api-client`'s `queryService` client + `queryService` SDK (https://github.com/keboola/ui, `packages/api-client`) — the formerly-standalone `query-service-api-js-sdk` repo is retired; `executeQuery`/`waitForJob`/pagination now live in `createQueryServiceSdk` (`@keboola/api-client/sdk/queryService`), composed on top of the plain `queryService` client (`@keboola/api-client/queryService`).
 
 Used when the app writes back to Storage tables via the Query Service (Snowflake projects with the **Storage Access** feature enabled). Provides authenticated `execute_query` against a permission-scoped workspace.
 
--- a/plugins/dataapp-developer/skills/dataapp-development/references/glossary.md
+++ b/plugins/dataapp-developer/skills/dataapp-development/references/glossary.md
@@ -76,11 +76,11 @@ See `references/kai-integration.md` for the embed patterns.
 
 Python SDK: https://github.com/keboola/query-service-api-python-sdk
 
-JS/TS SDK: https://github.com/keboola/query-service-api-js-sdk
+JS/TS: `@keboola/api-client`'s `queryService` client (https://github.com/keboola/ui, `packages/api-client`) — the formerly-standalone `query-service-api-js-sdk` repo is retired; its `executeQuery`/pagination behavior now lives here.
 
 Used when the app writes back to Storage tables via the Query Service (Snowflake projects with the **Storage Access** feature enabled). Provides authenticated `execute_query` against a permission-scoped workspace.
 
-First-class SQL helpers (`SQL.literal()`, `SQL.ident()`, `sql.format()`) are in development in both SDKs. Until released, use the manual sanitization pattern shown in `references/storage-access.md` (allowlists for strings, type-coercion for numbers).
+First-class SQL helpers (`SQL.literal()`, `SQL.ident()`, `sql.format()`) are in development for the Python SDK. Until released, use the manual sanitization pattern shown in `references/storage-access.md` (allowlists for strings, type-coercion for numbers).
 
 Install:
 
@@ -89,7 +89,7 @@ Install:
 uv add keboola-query-service
 
 # JS
-npm install @keboola/query-service
+npm install @keboola/api-client
 ```
 
 ### keboola-streamlit (helper utilities)
--- a/plugins/dataapp-developer/skills/dataapp-development/TODO.md
+++ b/plugins/dataapp-developer/skills/dataapp-development/TODO.md
@@ -36,7 +36,7 @@ Sections we know are incomplete because the underlying pattern isn't firm yet. M
 
 - **`storage-access.md` §Data access management — PLACEHOLDER.** Per-user / row-level data access control. No documented pattern; internal apps diverge. Cross-referenced from `authentication.md`.
 - **`python-js-apps.md` §Deployment via MCP — PLACEHOLDER.** Fill in once `modify_streamlit_data_app` covers Python/JS.
-- **SQL helpers in Query Service SDKs.** Once `SQL.literal()` / `SQL.ident()` / `sql.format()` ship in `keboola-query-service` (Py) and `@keboola/query-service` (JS), replace the manual sanitization patterns in `storage-access.md` §SQL injection with SDK-driven examples.
+- **SQL helpers in Query Service SDKs.** Once `SQL.literal()` / `SQL.ident()` / `sql.format()` ship in `keboola-query-service` (Py) and `@keboola/api-client`'s `queryService` client (JS), replace the manual sanitization patterns in `storage-access.md` §SQL injection with SDK-driven examples.
 - **Two Max Ottomansky suggestions from AI-3147 not yet picked up:**
   - Prebuilt JS apps — committing `dist/` to skip `npm install` / build on cold start. Worth a short subsection in `python-js-apps.md` once the deployment story is settled.
   - `KAI_TOKEN` secret workaround for embedding Kai chat without manual user token entry. Belongs in `kai-integration.md` once the contract with `kai-client` is firm.
--- a/plugins/dataapp-developer/skills/dataapp-development/references/duckdb-caching.md
+++ b/plugins/dataapp-developer/skills/dataapp-development/references/duckdb-caching.md
@@ -233,6 +233,19 @@ Use DuckDB when:
 
 You can combine both: DuckDB for the shared cache + `@st.cache_data` for derived per-session aggregations.
 
+## Verifying the cache is complete
+
+A cache silently holding a fraction of the table is worse than no cache — the dashboard stays fast and green while every KPI built on it is wrong. After each `refresh()`, compare `rowCount`/`_row_count` against the table's real row count from `get_table`'s `rows_count` field (or `SELECT COUNT(*)` against the source), and log/alert on a mismatch instead of trusting the pull silently succeeded:
+
+```javascript
+const expected = await getExpectedRowCount(); // e.g. mcp get_table().rows_count
+if (rowCount !== expected) {
+  console.error(`[duck] cache mismatch: pulled ${rowCount}, source has ${expected}`);
+}
+```
+
+This catches exactly the failure mode of pulling through a truncating query path (e.g. an unpatched SDK, or a hand-rolled query capped at a default page size) — the pull "succeeds" with a plausible-looking row count that's actually a silent truncation.
+
 ## Template
 
 Runnable starters live at:
--- a/plugins/dataapp-developer/skills/dataapp-development/references/storage-access.md
+++ b/plugins/dataapp-developer/skills/dataapp-development/references/storage-access.md
@@ -121,7 +121,7 @@ Use this when:
 Two paths to call the workspace:
 
 - **MCP-injected `query_data`** — when `modify_streamlit_data_app` is involved, a `query_data(sql) -> pd.DataFrame` function is dropped into the source code via the `{QUERY_DATA_FUNCTION}` placeholder. Use it as-is; don't roll your own.
-- **Query Service via the official SDK** — for Python/JS apps without MCP injection, call the Query Service API (`https://query.<stack>.keboola.com/api/v1/...`) using `keboola-query-service` (Python) or `@keboola/query-service` (JS/TS). The SDK handles submit + poll + paginate; you call `executeQuery({ branchId, workspaceId, statements })` and get back columns + rows.
+- **Query Service via the official SDK** — for Python/JS apps without MCP injection, call the Query Service API (`https://query.<stack>.keboola.com/api/v1/...`) using `keboola-query-service` (Python) or `@keboola/api-client`'s `queryService` client (JS/TS). The SDK handles submit + poll + paginate; you call `executeQuery(...)` and get back columns + rows.
 
 **On Snowflake, do NOT post to `/v2/storage/branch/<b>/workspaces/<w>/query`.** That older Storage API workspace-query endpoint returns `workspace.workspaceNotFound` 404s on Snowflake projects — use the Query Service instead. On BigQuery it does work and is a valid alternative (see "Alternative: Storage API workspace-query endpoint" below), but default to the Query Service on both backends.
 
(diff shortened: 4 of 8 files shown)


Matovidlo:
@MiroCillik can you check this whether its not in conflict with your PR https://github.com/keboola/ai-kit/pull/98 many thanks :pray:

MiroCillik:
Thanks for picking this up — the row-count verification section in `duckdb-caching.md` is exactly the guard I was hoping for in AI-3752, and switching to the maintained client is clearly the right direction.

Flagging three things I hit while checking this against my open [#98](https://github.com/keboola/ai-kit/pull/98) (which touches the same three docs). I tested against the published package, so if v26 is unreleased and adds all of this, most of the below is just "blocked until it ships" — please say so and I'll stop worrying.

### 1. `@keboola/api-client@^26.0.0` isn't published

```
$ npm view @keboola/api-client dist-tags
{ latest: '25.0.0' }

$ npm view @keboola/api-client versions --json | tail
[... '21.0.0', '23.0.0', '24.0.0', '25.0.0']

$ npm install @keboola/api-client@^26
npm error code ETARGET
npm error notarget No matching version found for @keboola/api-client@^26.
```

This one is load-bearing: `keboola-config/setup.sh` runs `npm install --omit=dev` at deploy, so with an unresolvable pin **the container never starts** — a harder failure than the truncation this fixes. Worth holding the merge until v26 is on npm (or pinning to whatever version actually carries the new API).

### 2. `executeQuery` doesn't exist on the published client

Running the template's new code against 25.0.0:

```
TypeError: client.executeQuery is not a function
```

The `./queryService` subpath resolves fine — that part's correct — but the client it returns exposes only primitives:

```
createQueryJob, cancelQueryJob, getQueryJob, getQueryResults, getQueryHistory, exportResults
```

So `executeQuery(branchId, workspaceId, { statements, transactional })` must be new in v26. If it isn't, the template and the `storage-access.md` snippet both need rewriting onto `createQueryJob` → `getQueryJob` → `getQueryResults`.

### 3. The "paginate" claim is still there, and still unverified

`storage-access.md` keeps the sentence AI-3752 was actually about, with only the package name changed:

> The SDK handles submit + poll + **paginate**; you call `executeQuery(...)` and get back columns + rows.

I can't confirm or refute it, because the method isn't in the published SDK. What I can see is that `getQueryResults(queryJobId, statementId, query?)` still derives its query params from the OpenAPI `/api/v1/queries/{queryJobId}/{statementId}/results` endpoint — the same offset/page-size surface the old SDK defaulted to 500 on. So unless v26's `executeQuery` loops internally, the claim stays false and the silent truncation just moves to a new package name.

Two suggestions regardless of what v26 does:

- **Verify it before restating it.** The original defect wasn't the 500-row cap — it was that the cap was silent *and* the docs promised otherwise. A quick check against a >500-row table would settle it. Happy to run it on my side (I have a 2412-row table wired up) the moment v26 is installable.
- **Make truncation loud in the template** as a belt-and-braces measure: `getQueryResults` returns `numberOfRows`, so `data.length < numberOfRows` → throw or warn. Then a future SDK regression can't silently produce wrong dashboards. This complements the `duckdb-caching.md` row-count check nicely — that one catches it at cache level, this catches it at query level.

Also `exportResults` looks like it may be the intended full-result path (CSV export, analogous to `kbagent workspace query --full`) — might be worth considering for bulk pulls like the DuckDB cache refresh, where you want the whole table in one shot rather than N pages.

### Heads-up on overlap with #98

#98 also edits `TODO.md`, `storage-access.md` and `troubleshooting.md`. I test-merged the branches: `storage-access.md` and `troubleshooting.md` merge cleanly, and there's **one conflict in `TODO.md`** where our edits are complementary rather than contradictory —

- mine marks the `python-js-apps.md` §Deployment via MCP PLACEHOLDER as done (it removes it)
- yours renames the JS SDK in the SQL-helpers line

so the resolution is "keep both changes". Whichever merges second, I'm happy to do the rebase — no need for you to deal with it.

Matovidlo:
Thanks — all three findings check out, and #2 was worse than it looked. Fixed:

**1. `@keboola/api-client@^26.0.0` unpublished:** confirmed and worse than the pin itself suggested — the pending release (`keboola/ui#7866`, the changesets bot's Version Packages PR) shows the next version will be `27.0.0`, not `26.0.0`. `^26.0.0` would never resolve even after release (caret excludes the next major-position bump). Fixed the pin to `^27.0.0`. **Marked this PR blocked in the description** — holding the merge until `@keboola/api-client@27.0.0`+ is actually on npm, per your recommendation.

**2. `executeQuery` doesn't exist on the client — and after review on `keboola/ui#8059` it never will.** Per Josef Biros's review there, `executeQuery`/`waitForJob` moved out of the `queryService` client entirely into a separate `createQueryServiceSdk` (matching the existing `sdks/storage` pattern in that repo) — the client stays a plain method collection. So the template/docs needed a second fix beyond just "wait for publish": `getSdk().executeQuery(...)` via `createQueryServiceClient` + `createQueryServiceSdk({ queryServiceClient })`, not `client.executeQuery(...)`. Updated the Node template and every JS/TS example in `storage-access.md`/`glossary.md`/`troubleshooting.md`.

**3. The "paginate" claim:** agreed it needs empirical proof, not just tests against a mock. `ui#8059` has 10 automated tests (multi-page merge, `maxRows` cap, job-failure, timeout, validation) but nothing against a real backend. I'd very much take you up on running it against your 2,412-row table (`in.c-dbt-beers.beers`) once v27 is installable — that's the check that actually closes AI-3752, not the code landing.

**Also added your loud-truncation suggestion** to the Node template's `runQuery`: asserts `result.data.length === result.numberOfRows` after `executeQuery` returns, throws otherwise. Cheap belt-and-braces given the auto-pagination is now real, but you're right it shouldn't rely on that alone.

Holding off on `exportResults` for the DuckDB cache refresh for now — good idea, but a separate change from this truncation fix; happy to pick it up as a follow-up if you want to file it.

Re #98 overlap: noted, and thanks for already test-merging — I'll let you drive that rebase as you offered, whichever lands second.

plugins/dataapp-developer/skills/dataapp-development/references/glossary.md
```diff
@@ -76,11 +76,11 @@ See `references/kai-integration.md` for the embed patterns.
...
-JS/TS SDK: https://github.com/keboola/query-service-api-js-sdk
+JS/TS: `@keboola/api-client`'s `queryService` client + `queryService` SDK (https://github.com/keboola/ui, `packages/api-client`) — the formerly-standalone `query-service-api-js-sdk` repo is retired; `executeQuery`/`waitForJob`/pagination now live in `createQueryServiceSdk` (`@keboola/api-client/sdk/queryService`), composed on top of the plain `queryService` client (`@keboola/api-client/queryService`).
 
 Used when the app writes back to Storage tables via the Query Service (Snowflake projects with the **Storage Access** feature enabled). Provides authenticated `execute_query` against a permission-scoped workspace.
 
-First-class SQL helpers (`SQL.literal()`, `SQL.ident()`, `sql.format()`) are in development in both SDKs. Until released, use the manual sanitization pattern shown in `references/storage-access.md` (allowlists for strings, type-coercion for numbers).
+First-class SQL helpers (`SQL.literal()`, `SQL.ident()`, `sql.format()`) are in development for the Python SDK. Until released, use the manual sanitization pattern shown in `references/storage-access.md` (allowlists for strings, type-coercion for numbers).
```
MiroCillik, line 83: Toto by som tiez mohol dotiahnut dakedy 🙈

MiroCillik:
Re-reviewed `37043094`. All three points addressed, and you were right to correct me on the version — `^26.0.0` would indeed never have resolved even post-release, since caret excludes the next major-position bump. Good catch.

I could verify more this round than last, because the Python half is installable today. Results below.

## Verified working

**The truncation guard is not vacuous — I confirmed its premise.** The assertion rests on `numberOfRows` reporting the *full* result size rather than the returned page; if it were the page count the check would always pass. Tested against the old SDK on a known-truncated result (2,412-row table, `in.c-dbt-beers.beers`, project 40):

```
data.length   : 500
numberOfRows  : 2412
rowsAffected  : 0
```

So `numberOfRows` is the true total, and **your guard would have caught the original bug**. Worth keeping regardless of what v27 does.

**The Python half is a genuine fix, verified end-to-end.** Same table, same workspace, only the pin changed:

| `keboola-query-service` | rows returned | verdict |
|---|---|---|
| `0.2.0` (previous pin) | **500** / 2412 | truncated |
| `0.3.0` (this PR) | **2412** / 2412 | paginates |

`0.3.0` is on PyPI, so this ships today. It also settles the open question in AI-3752 about whether the Python SDK shared the defect — **it did**, meaning the Streamlit template has been silently truncating too. Nice catch bumping it.

**Also confirmed:** no `@keboola/query-service` references remain anywhere in the plugin — fully repointed, docs and templates consistent with the client+SDK split.

## Still blocked (as you noted)

`@keboola/api-client` latest on npm is still `25.0.0`, so `^27.0.0` doesn't resolve and `createQueryServiceSdk` can't be verified. Your commit message already says the PR stays blocked until release — agreed, and no action needed beyond waiting. Happy to re-run the JS side against the 2,412-row table the moment v27 lands.

## One thing I'd still change: the JS docs read as present tense with no version gate

`storage-access.md` still asserts, unqualified:

- L124 — *"The SDK handles submit + poll + paginate; you call `executeQuery(...)`…"*
- L205 — *"The SDKs handle the submit-job → poll-status → paginate-results dance internally. **Don't hand-roll that**"*

For **Python** both are now true and verified. For **JS**, with what's on npm today, an agent following this gets the worst of both: `executeQuery` doesn't exist so the documented call throws, and it's told not to hand-roll paging — which is currently the only thing that works. The skill is read by agents against whatever is installed, not against the release schedule.

Suggest gating those two statements, e.g. *"JS/TS: requires `@keboola/api-client` ≥ 27; on earlier versions page `getQueryResults` manually"* — or if you'd rather not carry the caveat, holding the JS doc changes until v27 ships and merging the Python bump plus the `duckdb-caching.md` guard now. The Python half is verified and independently valuable; no reason for it to wait on the npm release.

**Minor:** the guard skips when the field is absent —

```js
if (result.numberOfRows !== undefined && result.data.length !== result.numberOfRows)
```

If a future SDK stops returning `numberOfRows`, the check silently no-ops and looks like it's still protecting you. It's populated in practice (2412 above), so low severity, but a `console.warn` on the `undefined` branch would make the guard's own inapplicability visible.

## Overlap with #98

Unchanged from last round: `storage-access.md` and `troubleshooting.md` merge clean, one trivial `TODO.md` conflict where our edits are complementary (mine marks the `python-js-apps.md` PLACEHOLDER done, yours renames the JS SDK in the SQL-helpers line — resolution keeps both). Still happy to take the rebase whichever lands second.

Matovidlo:
`@keboola/api-client@27.0.0` is now on npm (`npm view @keboola/api-client dist-tags` → `latest: 27.0.0`), so this PR's blocker is resolved. Removed the "Blocked" banner from the description.

I don't have KBC credentials in this environment to run the live >500-row check, but I installed `27.0.0` directly and verified statically:

- `@keboola/api-client/queryService` exports `createQueryServiceClient`, `@keboola/api-client/sdk/queryService` exports `createQueryServiceSdk` — matches `api/keboola-client.js`'s imports exactly.
- Read the built SDK source: `executeQuery` → `createQueryJob` → `waitForJob` → `fetchAllResultPages`, which loops on `offset`/`pageSize` calling `getQueryResults` until `allData.length >= numberOfRows`. The auto-pagination claim in `storage-access.md` holds for real, it's not just a renamed package.
- `numberOfRows` in the response is the *total* row count (not the page size), confirmed by @MiroCillik's earlier truncation test against the old SDK (500/2412). So `runQuery`'s guard (`result.data.length !== result.numberOfRows`) is non-vacuous — it would have caught the original bug.

@MiroCillik — whenever you get a chance, re-running your `in.c-dbt-beers.beers` (2412-row) check against v27 would still be worth doing as the final live confirmation before merge.

MiroCillik:
Re-ran the check now that `@keboola/api-client@27.0.0` is on npm. **It paginates — the JS side is fixed.**

I ran this PR's `templates/nodejs-app/api/keboola-client.js` **verbatim** (fetched from the PR head, not paraphrased), against the same 2,412-row table as before, so this exercises the shipped code path — client + SDK construction, the new `executeQuery(branchId, workspaceId, {...})` signature, and the truncation guard — not just the SDK in isolation.

```
$ grep api-client package.json
    "@keboola/api-client": "^27.0.0",
$ npm install && <read installed version>
installed version: 27.0.0
  ./queryService         present=True
  ./sdk/queryService     present=True

$ node run.mjs          # imports runQuery from the PR's keboola-client.js unmodified
env: branch=15 workspace=1833354510 token=true
rows returned : 2412  (in 2338ms)
true total    : 2412
sample row    : {"id":1436,"name":"Pub Beer"}
```

For contrast, the same query on the old path returned **500**. So `executeQuery` now loops internally, and the *"handles submit + poll + paginate"* claim in `storage-access.md` is accurate as written — no version caveat needed once this merges.

## The guard still has teeth

Worth confirming explicitly, since it would fail open rather than closed: the assertion is skipped when `numberOfRows` is `undefined`, so if v27 had dropped the field the guard would silently stop protecting anything. It hasn't:

```
data.length    : 2412
numberOfRows   : 2412 (type number)
result keys    : status, data, columns, rowsAffected, numberOfRows, message
```

Populated as a number, and it didn't false-positive on a complete result. Combined with the earlier finding that `numberOfRows` reports the *full* total rather than the returned page (500 vs 2412 on a known-truncated result), the guard would catch a future regression.

## Both halves now verified

| Side | Pin | Result |
|---|---|---|
| Python | `keboola-query-service>=0.3.0` | 2412/2412 — and `0.2.0` gave 500/2412, so the bump is the fix |
| JS | `@keboola/api-client@^27.0.0` | 2412/2412 — `executeQuery` via `createQueryServiceSdk` |

Both blockers from my earlier review are cleared: the pin resolves, and `executeQuery` exists where the template expects it. Nothing outstanding from my side — LGTM.

One note for whoever merges: this needs `TODO.md` conflict resolution against #98 (mine marks the `python-js-apps.md` PLACEHOLDER done, yours renames the JS SDK in the SQL-helpers line — keep both). #98's scope shrank since I last mentioned this, but that one conflict remains. Happy to rebase mine after you land.

MiroCillik (approved):
LGTM now :)

MiroCillik:
So we can now merge this, and then I will rebase the #98

PR #111, opened 2026-09-15
Kai builds into the existing app instead of creating a second one (AJDA-3263)

natocTo:
Closes [AJDA-3263](https://linear.app/keboola/issue/AJDA-3263/).

## The measurement

On project 1255 an empty Python/JS app with a Keboola-managed Git repository already existed, the UI opened Kai on its page, and Kai got a plain build request. Kai named the right app from the page path — its own words were "I'll build this into your existing app" — and then called `modify_python_js_data_app` **without** `configuration_id`, which creates. Two apps: the pre-made one untouched and unnamed, holding a repo nobody will ever use, and Kai's own carrying the work.

## Why the fix is shaped this way

The obvious framing, "check whether an app exists before you create one", is the one the failing agent already satisfied. It resolved correctly and said so out loud, then typed the wrong argument anyway.

The actual trap is that **creating a prod app and creating a draft are the same call with the same empty `configuration_id`**. Only `parent_configuration_id` separates them. So the rule is an invariant on the argument, which still holds when the model has convinced itself it is on the right app:

> Never send an empty `configuration_id` without either `parent_configuration_id`, or a sentence in your reply saying you are creating a new resource and why no existing one fits.

Generalised in `SKILL.md` hard rule 10 so it covers `modify_streamlit_data_app` too, and kept short enough to survive a skim.

One simplification worth stating: there is only **one** flow, a draft off a prod app. Creating a prod app is the rare branch. Whether the prod app is empty or already built changes exactly one thing, the git command that starts the branch, because a fresh managed repo has no `origin/main` yet.

## What changed

**New** `references/python-js-prod-and-drafts.md` — invariant and the three-way argument table up front, then the prod/drafts model, the resolution order, and the draft flow end to end.

**Gated the create step in every copy-pasteable snippet that had it bare.** Prose next to a fenced block does not help; an agent lifts the block. This covers `keboola-git` SKILL.md and `managed-git.md`, `mcp-data-app/reference/deploy.md`, and `/keboola-git-copy`, which defaulted to creating whenever `--config` was omitted and is a procedure an agent executes literally.

**Dropped stale claims.** The skill still said Python/JS had no MCP support and to fall back to a customer-provided GitHub — in `choosing-app-type.md` (the first hop for "build a new app"), in `deployment-paths.md` Path A and its "How to choose" table, which contradicted its own body 130 lines earlier, and in the `python-js-apps.md` PLACEHOLDER section. Also fixed four pointers to `dataapp-dev` / `dataapp-deployment`, skills that no longer exist.

Versions bumped per CLAUDE.md: dataapp-developer 1.5.1 → 1.6.0, keboola-git 1.0.1 → 1.1.0, marketplace 1.13.0 → 1.14.0.

## Before the keboola/ui parenthetical comes out

`DataAppsEmptyState.tsx` currently appends "(Build into the app already open here — configuration `<id>` — rather than creating a new one.)" to the user's first message. Nothing here depends on it, but **don't delete it yet**: the Kai addendum vendored into the sandbox carries its own copy of the create step and will keep overriding this skill until it gets the same rule. I have not been able to locate that addendum — it is in neither `keboola/ai-kit` nor `keboola/ai-kit-internal`.

## Verification and what is not verified

Tier 0 lint passes (331 passed, 32 skipped). The flow is written from the `keboola/mcp-server` tool contract, not from a live run — `TODO.md` records that, plus two specific claims to confirm on the first real pass: what `deploy_data_app(mode="dev")` does about authentication for the preview, and whether a dev deployment follows the branch or stays pinned by commit locking.

Worth knowing: the repo only has trigger evals, and this failure was not an activation failure — the skill was reachable and the agent still sent the wrong arguments. So CI cannot catch a regression here. That is noted in `TODO.md`, and it is why the invariant lives in always-loaded `SKILL.md` rather than only in the reference.


natocTo:
Na novém API to bude zase jinak, ale mělo by to asi nějak popisovat ted ten současný stav.

cjayyy (approved):

PR #114, opened 2026-09-18
docs(dataapp-developer): no-auth is prod-only, rejected on drafts

cjayyy:
## Summary
Companion to keboola/mcp-server#713 (AJDA-3237): the MCP now rejects `authentication_type='no-auth'` on Python/JS **drafts** (create with `parent_configuration_id`, update of an `isDraft` config). `no-auth` stays valid on prod Python/JS apps and on Streamlit apps, and only when the user explicitly asked for a public app.

`dataapp-development` skill references:
- `authentication.md` — overview names both `modify_streamlit_data_app` / `modify_python_js_data_app`; Streamlit `## MCP defaults` gains the "only for an explicitly public app" caveat; new `## Python/JS: no-auth is prod-only, never on a draft` section (draft rejection, preview authenticates on top of configured auth per the apps-proxy `kai-preview` design, set `no-auth` on prod at promote time); OIDC bootstrap deploys with `basic-auth` (no longer `none`).
- `python-js-prod-and-drafts.md` — draft-create example passes `authentication_type="basic-auth"`; preview section: a non-loading preview is never fixed by `no-auth`; keeps the separate "`mode="dev"` unconfirmed on a live stack" fallback note.
- `deployment-paths.md` — Streamlit example comment: `"no-auth"` only when the user asked to make the app public.
- `TODO.md` — preview-auth listed as a claim to confirm on the live run (design-backed, not asserted as settled).

Plugin bumped 1.6.0 → 1.6.1 (`plugin.json`, `marketplace.json`, plugin README).

Link to Devin session: https://app.devin.ai/sessions/9be8db5f499647829f274885247e3596
Open in Devin Desktop: https://app.devin.ai/desktop/session/9be8db5f499647829f274885247e3596?variant=devin
Requested by: @cjayyy


MiroCillik (commented):
Docs-only review. Version bumps are consistent across `plugin.json`, the `marketplace.json` entry, and the plugin README (1.6.0 → 1.6.1); the root README correctly needed no change since no skills or commands were added.

5 findings, none blocking. Four are inline; the two below can't be anchored because they point at a file this PR doesn't touch.

---

**medium — `references/python-js-prod-and-drafts.md:122`**

This is the file the draft/preview flow actually routes through, and it's where the failure this PR guards against happens. §5 "Deploy the draft as a preview" tells the agent that if `deploy_data_app(mode="dev")` is rejected it should "drop `mode` and deploy the draft plainly", and says nothing about auth.

The new rule lives only in `authentication.md`, which `SKILL.md:48` routes to only for "Securing the app (login, SSO, OAuth)" — a preview-won't-load debugging session never loads that file. An agent whose preview fails to open reaches for `authentication_type="no-auth"` exactly as before, because the file it is reading never mentions it.

The preview-auth sentence (or a pointer to it) belongs in §5 and/or the §2 draft-create snippet.

---

**Unverified claim worth confirming before merge**

The new rationale "a draft inherits the prod app's data access" isn't corroborated anywhere else in the skill — `python-js-prod-and-drafts.md:151` lists `storage` as a per-config field settable on update, which suggests a draft has its own storage block rather than inheriting one. The rationale may come from `keboola/mcp-server#713`, which I couldn't read (GitHub MCP server failed to connect).

The security conclusion holds either way; it's the stated reason that may need rewording.

plugins/dataapp-developer/skills/dataapp-development/references/authentication.md
```diff
@@ -74,6 +74,14 @@ When using `modify_streamlit_data_app` via MCP:
 
 Acceptable values: `"no-auth"`, `"basic-auth"`, `"default"`.
 
+`"no-auth"` is only for an app the user has explicitly asked to make public, and
```
MiroCillik: **low — section placement**

This paragraph is inserted inside `## MCP defaults`, which line 70 explicitly scopes ("When using `modify_streamlit_data_app` via MCP:") and whose immediately preceding line 75 is the Streamlit tool's acceptable-values list. The paragraph itself describes `modify_python_js_data_app` draft rejection — a tool with a prod/draft model that `modify_streamlit_data_app` does not have.

A reader skimming the section will either attach the draft rejection to the Streamlit tool, or skip the paragraph as Streamlit-specific and miss it for Python/JS.

Suggest giving it its own `##` heading (or a Python/JS subsection). If you do, the file's opening line 8-9 ("configured … via `modify_streamlit_data_app`") should then name both tools.

plugins/dataapp-developer/skills/dataapp-development/references/deployment-paths.md
```diff
@@ -99,7 +99,7 @@ import streamlit as st
 # ... your app code using query_data(sql)
 """,
     packages=["pandas", "plotly"],
-    authentication_type="basic-auth",  # or "no-auth" / "default" on update
+    authentication_type="basic-auth",  # "default" on update; "no-auth" only on a prod app the user asked to make public
```
MiroCillik: **low — prod/draft distinction doesn't apply here**

This is a `modify_streamlit_data_app` call, but drafts (and therefore the prod-vs-draft distinction) are a Python/JS-only concept per `python-js-prod-and-drafts.md`. Every Streamlit app is "the prod app".

As written, "only on a prod app" implies Streamlit apps also have drafts where `no-auth` is forbidden, which will send a reader looking for a draft flow that doesn't exist.

Either drop the prod clause here:

```suggestion
    authentication_type="basic-auth",  # "default" on update; "no-auth" only when the user asked to make the app public
```

or scope the prod-only wording to Python/JS.

plugins/dataapp-developer/skills/dataapp-development/TODO.md
```diff
@@ -51,7 +51,7 @@ The skill has been validated end-to-end in three sessions, but not against every
...
-- **Prod-app-and-drafts flow** (`references/python-js-prod-and-drafts.md`) — written from the `keboola/mcp-server` tool contract and the AJDA-3263 measurement, not yet driven end to end from a fresh Apps-page app (rename prod → draft → push → `mode="dev"` preview → merge → prod deploy). Two claims to confirm on that run: what `deploy_data_app(mode="dev")` does about authentication for the preview, and whether a dev deployment follows the branch or stays pinned by commit locking.
+- **Prod-app-and-drafts flow** (`references/python-js-prod-and-drafts.md`) — written from the `keboola/mcp-server` tool contract and the AJDA-3263 measurement, not yet driven end to end from a fresh Apps-page app (rename prod → draft → push → `mode="dev"` preview → merge → prod deploy). One claim to confirm on that run: whether a dev deployment follows the branch or stays pinned by commit locking. (Preview authentication is settled: the Kai preview iframe authenticates via an apps-proxy `kai-preview` cookie layered on top of the app's configured auth — see `keboola/ui` `apps/kai-agent/docs/data-apps-architecture.md` and AJDA-3237 — so `mode="dev"` never needs auth disabled.)
```
MiroCillik: **low — "settled" sits awkwardly next to the rest of the bullet**

This now declares preview authentication settled and drops it as a live-test item, while the same bullet still says the whole `mode="dev"` preview flow has never been driven end to end — and `python-js-prod-and-drafts.md:122` still tells the agent `mode="dev"` "is not yet confirmed against a live stack … if the call rejects it, drop `mode`".

So the skill now asserts unconditionally what `mode="dev"` does about authentication (`authentication.md`, new lines 83-84) while elsewhere warning the argument may not exist at all. The evidence cited is a design doc in `keboola/ui`, not an observation.

Suggest either hedging the `authentication.md` sentence ("per the apps-proxy `kai-preview` design") or keeping this item open until the same run that settles the branch-pinning question.

plugins/dataapp-developer/skills/dataapp-development/references/authentication.md
MiroCillik, line 1: **medium — line 55 (outside this diff, so anchored at file level)**

The OIDC bootstrap recipe still reads:

> 1. Deploy the app once (with `none` or `basic-auth`) to obtain the `dataAppId`.

This is the same file the PR just amended to say `"no-auth"` is *only* for an app the user explicitly asked to make public. An agent setting up OIDC follows this numbered list literally, picks `none`, and either publishes a Storage-reading app to the open internet for the duration of the IdP setup, or — on a Python/JS draft — hits the new upstream rejection with no hint why.

`basic-auth` alone satisfies the step. Suggest dropping the `none` option:

> 1. Deploy the app once (with `basic-auth`) to obtain the `dataAppId`.

MiroCillik (approved):
All five review findings are addressed in 7d4a3b6, and both CI tiers pass.

- **OIDC bootstrap** (`authentication.md:55`) — `none` dropped, `basic-auth` only.
- **Drafts flow** (`python-js-prod-and-drafts.md`) — fixed in both places: §5 now says the fix for a preview that won't load is never `authentication_type="no-auth"`, and the §2 draft-create snippet carries the inline comment. This was the important one, since a preview-debugging session never loads `authentication.md`.
- **Section placement** — the rule now has its own `## Python/JS: no-auth is prod-only, never on a draft` section; `## MCP defaults` is trimmed back to what's true for Streamlit, and the file's opening line names both tools.
- **`deployment-paths.md:102`** — prod/draft clause dropped, since Streamlit apps have no drafts.
- **`TODO.md:54`** — back to "two claims to confirm", hedged as design-backed rather than observed, plus the useful addition that `"no-auth"` isn't the answer either way.

I also verified the upstream behaviour against keboola/mcp-server#713 (merged), which I couldn't read during the first pass. It confirms the docs as written: `ValueError` on the create path when `parent_configuration_id` is set, `ValueError` on the update path via `_is_draft_config(...)`, `no-auth` still available for prod Python/JS and for Streamlit, and the kai-preview cookie being additive to the configured auth.

One note for the record: I'd flagged "a draft inherits the prod app's data access" as uncorroborated. The upstream PR description uses that exact phrasing, so the flag was over-cautious — and the rewrite ("runs the prod app's code in the same project with its own Storage workspace and token") is more precise anyway. No change needed.

Version correctly stays at 1.6.1 — docs-only edits to an existing skill, already bumped earlier in this PR.

